PR #65, opened 2021-06-11
Yoma-432  - refactor(AppNavigation): navigate to user dash landing screen on successful authentication

sciphergfx:
## Work Done
navigate to user dash landing screen on successful login
added test for selector

## Steps to Test
Sign in with user credentials
    
## Screenshots
https://user-images.githubusercontent.com/6403720/121633746-aaccb100-ca83-11eb-8237-2a6f9ac7adee.mp4


cameronolivier (changes requested):

cameronolivier (changes requested):

cameronolivier (changes requested):

cameronolivier (changes requested):

cameronolivier (changes requested):

cameronolivier (changes requested):
one small one

cameronolivier (dismissed):

cameronolivier (changes requested):
It seems there's some code from PR#64 that's found its way in here - please clean that up - the rest looks good.

cameronolivier (changes requested):

cameronolivier (changes requested):

src/modules/Auth/Auth.selector.test.ts
```diff
@@ -0,0 +1,68 @@
+import { INITIAL_STATE } from './Auth.reducer'
+import * as SUT from './Auth.selector'
+
+describe('modules/Auth/Auth.selector', () => {
+  describe('selectAuthState', () => {
+    it('should return auth property of the root state', () => {
```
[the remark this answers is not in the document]
cameronolivier: but you're testing that the auth state itself returns the expected values - not that auth exists
So you're testing that if you get the auth endpoint that it 
1. looks like the INITIAL_STATE for the auth property if there is no data and 
2. looks like it should if it's populated (with mock data obviously for the test)

src/modules/Auth/Auth.selector.ts
```diff
@@ -0,0 +1,7 @@
+import { createSelector } from '@reduxjs/toolkit'
+import { isNil, complement, propOr } from 'ramda'
+import { RootState } from 'redux/redux.types'
+
+export const selectAuthState = (state: Partial<RootState>) => state.auth
```
cameronolivier: state shouldn't be 'partial' rootstate, but the entire rootstate. 
```js
export const selectAuthState = (state: RootState) => state.auth
```

src/modules/Auth/Auth.selector.test.ts
```diff
@@ -0,0 +1,75 @@
...
+          token: 'USER_TOKEN',
+          expiresAt: 'EXPIRY_DATE',
+        },
+      }
+      // when ... we call the selector
+      const result = SUT.selectAuthState(state)
+      // then ... should return result as expected
+      expect(result).toEqual(state.auth)
+    })
+    it('should return the default auth state', () => {
```
cameronolivier, line 19: OPTIONAL: small convention thing, but personally I prefer default/empty state tests to come first. But not a huge issue if it remains like this.

cameronolivier (dismissed):
One tiny change and then an optional suggestion. Nearly there! :) 

Considering the rush, happy to get this in - and we can do these changes in whatever you're working on, once you merge develop into your branch (if that makes sense)

cameronolivier (approved):

PR #66, opened 2021-06-12
Yoma-422 - refactor(profile): user logout feature

sciphergfx:
## Work Done
Added user logout 
Fixed lint issues


cameronolivier (changes requested):

cameronolivier (changes requested):

cameronolivier (changes requested):

cameronolivier (changes requested):

cameronolivier (changes requested):

cameronolivier (changes requested):

src/modules/App/App.middleware.test.ts
```diff
@@ -0,0 +1,29 @@
...
+    it('should handle app reset action accurately', async () => {
+      const create = createMiddlewareMock(jest)
+
+      // given ... the resetAppData action is fired
+      const action = resetAppData()
+      const { store, invoke } = create(SUT.appResetFlow)
+
+      await invoke(action)
+      // ... we validate that our actions were triggered
+      expect(store.dispatch).toBeCalled()
```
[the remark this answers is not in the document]
cameronolivier: we should be testing that the actions we're expecting to be fired are fired, so it's fine to import in this case.

src/modules/Profile/Profile.container.ts
```diff
@@ -0,0 +1,16 @@
...
+const mapStateToProps = null
+const mapDispatchToProps = (dispatch: Dispatch) => {
+  return {
+    onLogoutUser: () => {
+      dispatch(AppActions.resetAppData())
+    },
+  }
+}
+
+export default connect(mapStateToProps, mapDispatchToProps)(Profile as any)
```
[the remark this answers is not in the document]
cameronolivier: not a huge issue, but
```js
(Profile as any)
```
shouldn't be necessary

src/redux/middleware.ts
```diff
@@ -1,9 +1,9 @@
 import { concat } from 'ramda'
 import { Middleware } from 'redux'
 
-import api from '../api'
-import { middleware as apiMiddleware, apiConfig } from '../api'
+import api, { apiConfig, middleware as apiMiddleware } from '../api'
 import { prepareApiRequest } from '../api/api.utils'
+import * as appMiddleware from '../modules/App/App.middleware'
```
cameronolivier: can you export the same as the other modules - from the index, please.

cameronolivier (changes requested):

src/modules/Profile/Profile.container.ts
```diff
@@ -0,0 +1,16 @@
...
+import { Dispatch } from 'redux'
+
+import * as AppActions from '../App/App.reducer'
+import Profile from './Profile'
+
+const mapStateToProps = null
+const mapDispatchToProps = (dispatch: Dispatch) => {
+  return {
+    onLogoutUser: () => {
+      dispatch(AppActions.resetAppData())
```
cameronolivier: This should dispatch the `AuthActions.logoutUser` action, not the `resetAppData` action.

cameronolivier (changes requested):

src/modules/Profile/Profile.container.ts
sciphergfx: I did it like this previously before this change

src/modules/Auth/Auth.reducer.test.ts
```diff
@@ -35,4 +35,18 @@ describe('modules/Auth/Auth.reducer', () => {
       expect(result).toEqual(credentials)
     })
   })
+  describe('clearAuth', () => {
+    it('should clear auth state', () => {
+      // give ... there are no credentials in state
```
cameronolivier, line 40: should be 'given' I think that's maybe a typo I started.
and then it should either say `given ... there is data in state` (or simply remove the description, and just have `given ....`

Right now it says given there are no credentials - which definitely isn't what we're testing.

cameronolivier (changes requested):
There's one or 2 still outstanding here, and then the on comment now - but small things, just tidying up.

cameronolivier (dismissed):

cameronolivier (approved):

PR #68, opened 2021-06-18
Yoma-435 - refactor(Auth) - registration using api redux middleware

sciphergfx:
## Work Done
- registration using api redux middleware

## Steps to Test 
run the tests.


cameronolivier (changes requested):
Some changes.

cameronolivier (changes requested):

src/modules/Auth/Auth.utils.ts
```diff
@@ -1,5 +1,5 @@
 import { path, pathOr, pick, pipe } from 'ramda'
 
 export const selectCredentialsFromLoginPayload = pipe(path(['payload', 'data', 'data']), pick(['token', 'expiresAt']))
-
 export const selectRefreshTokenFromLoginPayload = pathOr('', ['payload', 'data', 'data', 'refreshToken'])
+export const selectLoginCredentialsFromRegistration = pipe(pick(['email', 'password']))
```
cameronolivier: no need for pipe here as you're only using a single util - so can be called on its own.

cameronolivier (changes requested):

src/modules/Auth/Auth.middleware.ts
```diff
@@ -100,9 +106,11 @@ export const authRegistrationFlow =
 
     // TODO: Abstract the api calls into a single api middleware
     if (authRegistration.match(action)) {
+      const credentials = selectLoginCredentialsFromRegistration(action.payload)
       await api.auth
```
cameronolivier: The task here was to make use of the api redux middleware - you're using the api module here. Please refactor the registration to use the middleware - see the Login flow as an example.

So as it stands, this isn't doing what is expected by the story.

Once we have that implemented, we should look at adding the login on success part of it (I think the way we're going to need to do that is through temporarily storing the login credentials in state, and then removing them again.

cameronolivier (changes requested):

src/modules/Auth/Auth.types.ts
```diff
@@ -13,6 +13,8 @@ export interface AuthRefreshTokenResponse {
 export interface AuthCredentialsResponse {
   token: string
   expiresAt: string
+  email?: string
+  password?: string
```
sciphergfx, line 17: This is temporary, I think we can add the credentials here to the incoming user module.

tests/tests.utils.ts
```diff
@@ -1,8 +1,8 @@
 import { Action, Middleware } from 'redux'
 
-export const createMiddlewareMock = (jest: any) => (middleware: Middleware) => {
+export const createMiddlewareMock = (jest: any, state?: {}) => (middleware: Middleware) => {
```
sciphergfx, line 3: Added this to be able to add mock state data in test

src/modules/Auth/Auth.middleware.test.ts
```diff
@@ -220,69 +221,63 @@ describe('modules/Auth/Auth.middleware', () => {
...
+          ...INITIAL_STATE,
+          email: 'EMAIL',
+          password: 'PASSWORD',
+        },
+      }
+      const create = createMiddlewareMock(jest, state)
       // @ts-ignore
       const response = defaultUserRegistrationResponseData
       const action = authRegistrationSuccess(response)
       const mockNotification = jest.fn()
```
cameronolivier, line 277: can leave this for now - (So OPTIONAL) but functions should really be called Stubs (I think I've messed this up as well) - so should be notificationStub. 

We 'stub' functions and 'mock' data.

src/modules/Auth/Auth.middleware.ts
```diff
@@ -91,40 +98,44 @@ export const authLoginFailureFlow =
...
-          dispatch(authRegistrationSuccess(response))
-        })
-        .catch((error: AuthRegistrationFailureResponse) => {
-          dispatch(authRegistrationFailure(error))
-        })
+      const credentials = selectLoginCredentialsFromRegistration(action.payload)
+      dispatch(
+        ApiActions.apiRequest(
+          mergeRight(ApiAuthConstants.REGISTER_CONFIG, {
+            isTokenRequired: false,
```
cameronolivier: Optional (we can always refactor later) but this isn't required as it's set as a default in the API auth config

src/modules/Auth/Auth.middleware.ts
```diff
@@ -91,40 +98,44 @@ export const authLoginFailureFlow =
...
 
     return result
   }
 
 export const authRegistrationSuccessFlow =
   ({ notification }: { notification: typeof showSimpleMessage }): Middleware =>
-  _store =>
+  ({ getState, dispatch }) =>
   next =>
   async action => {
```
cameronolivier: this action doesn't need to be async

src/modules/Auth/Auth.middleware.ts
```diff
@@ -134,8 +145,8 @@ export const authRegistrationFailureFlow =
...
     const result = next(action)
-
     if (authRegistrationFailure.match(action)) {
+      NavigationActions.navigate(AuthNavigationRoutes.Register)
       // TODO: this should be handled by the notification module
       notification('danger', 'Error', action.payload)
```
cameronolivier, line 149: we should dispatch a failure action (even if we aren't handling it right now)

src/modules/Auth/Auth.types.ts
cameronolivier: That or we could add it to a `temp` state endpoint that's blacklisted from persisted state? - not sure if that's a bad idea or not...

tests/tests.utils.ts
cameronolivier: That's cool :) 

you can also do something like this in the test, but yours is actually a cleaner solution
```js
store.getState = jest.fn(() => stateMock)
```

cameronolivier (changes requested):
a couple of changes

src/modules/Auth/Auth.middleware.ts
sciphergfx: noted @cameronolivier

sciphergfx: Please elaborate

sciphergfx: Sorry for replying late here, tried commenting earlier. Could you confirm which failure action here? As this middleware is handling authRegistrationFailure action

cameronolivier: Right. Sorry- misread :(

cameronolivier (dismissed):

cameronolivier (dismissed):

cameronolivier (approved):

PR #70, opened 2021-06-20
YOMA-434 - refactor(Auth): add sso registration

sciphergfx:
## Work Done
- Added facebook and google registration and sign in SSO

## Steps to Test
Open the app,  tap facebook button then follow prompts (Google auth provider is still an issue on the backend but it's been configured on the app)


src/api/api.middleware.ts
```diff
@@ -17,10 +17,12 @@ export const apiFlow =
       const { onSuccess, onFailure, apiArgs } = prepArgs(state, action)
       await api(apiArgs)
         .then((response: any) => {
+          console.log('response', response)
```
cameronolivier: Can you remove the log

src/api/api.middleware.ts
```diff
@@ -17,10 +17,12 @@ export const apiFlow =
...
           const serializableResponse = omit(['config', 'request'], response)
           dispatch(onSuccess(serializableResponse))
         })
         .catch((error: any) => {
+          console.log('error', error)
```
cameronolivier: and here

src/components/SocialLogin/SocialLogin.tsx
```diff
@@ -1,42 +1,18 @@
...
-      await GoogleSignin.configure()
-      await GoogleSignin.hasPlayServices()
-      const userInfo = await GoogleSignin.signIn()
-      console.log('User Info', userInfo)
-    } catch (error) {
-      console.log('User error', error)
-    }
-  }
+type Props = {
+  onLoginWithSocial: (authProvider: string) => void
```
cameronolivier: `authProvider` should be typed to `SocialVariants`.

src/components/SocialRegistration/SocialRegistration.tsx
```diff
@@ -1,61 +1,18 @@
...
-        // play services not available or outdated
-      } else {
-        console.log('error', error)
-        // some other error happened
-      }
-    }
-  }
-
+type Props = {
+  onRegisterWithSocial: (authProvider: string) => void
```
cameronolivier: likewise here - should be:  `authProvider: SocialVariants`

src/components/SocialRegistration/SocialRegistration.tsx
```diff
@@ -1,61 +1,18 @@
-import { GoogleSignin, statusCodes } from '@react-native-google-signin/google-signin'
 import React from 'react'
```
cameronolivier: This and Login should be the same component- it's doing 100% the same thing in the component - we're just wiring it up to different (login or reg) on implementation.

src/modules/Auth/Auth.middleware.ts
```diff
@@ -40,6 +45,88 @@ export const authLoginFlow: Middleware =
...
+  async action => {
+    const result = next(action)
+    if (authSocialLogin.match(action)) {
+      try {
+        const authProvider = action.payload
+        const authdata = await socialAuth(authProvider)
+        const credentials = selectLoginCredentials(authProvider, authdata)
+        dispatch(authSocialLoginSuccess(credentials))
+      } catch (error) {
+        notification('danger', 'Error', error)
```
cameronolivier: I think we should dispatch a failure action here

src/modules/Auth/Auth.middleware.ts
```diff
@@ -12,13 +12,18 @@ import {
...
+  authSocialRegistration,
+  authSocialRegistrationSuccess,
   setAuthCredentials,
   setSecureRefreshToken,
   setSecureRefreshTokenFailure,
   setSecureRefreshTokenSuccess,
 } from './Auth.reducer'
 import { AuthRegistrationFailureResponse, AuthRegistrationSuccessResponse } from './Auth.types'
 import { selectCredentialsFromLoginPayload, selectRefreshTokenFromLoginPayload } from './Auth.utils'
+import { selectLoginCredentials, selectRegistrationCredentials } from './Social/Social.utils'
```
cameronolivier: these should be in the auth utils as Social auth is still part of auth.

cameronolivier: I don't think we need a separate Social module

src/modules/Auth/Auth.middleware.ts
```diff
@@ -40,6 +45,88 @@ export const authLoginFlow: Middleware =
...
+  ({ dispatch }) =>
+  next =>
+  action => {
+    const result = next(action)
+
+    if (authSocialLoginSuccess.match(action)) {
+      dispatch(
+        ApiActions.apiRequest(
+          mergeRight(ApiAuthConstants.LOGIN_SOCIAL_CONFIG, {
+            isTokenRequired: false,
```
cameronolivier: this is actually added to the default constants in ApiAuthConstants - so don't need to explicitly set it here.

src/modules/Auth/Auth.middleware.ts
```diff
@@ -40,6 +45,88 @@ export const authLoginFlow: Middleware =
...
+  async action => {
+    const result = next(action)
+    if (authSocialRegistration.match(action)) {
+      try {
+        const authProvider = action.payload
+        const authdata = await socialAuth(authProvider)
+        const credentials = selectRegistrationCredentials(authProvider, authdata)
+        dispatch(authSocialRegistrationSuccess(credentials))
+      } catch (error) {
+        notification('danger', 'Error', error)
```
cameronolivier: we should dispatch a failure action

src/modules/Auth/Auth.middleware.ts
```diff
@@ -40,6 +45,88 @@ export const authLoginFlow: Middleware =
...
+  ({ dispatch }) =>
+  next =>
+  async action => {
+    const result = next(action)
+
+    if (authSocialRegistrationSuccess.match(action)) {
+      dispatch(
+        ApiActions.apiRequest(
+          mergeRight(ApiAuthConstants.REGISTER_SOCIAL_CONFIG, {
+            isTokenRequired: false,
```
cameronolivier: can remove isTokenRequired as it's set in the default for the auth api

src/modules/Auth/Social/Social.config.ts
```diff
@@ -0,0 +1,40 @@
...
+import { AccessToken, LoginManager, Profile } from 'react-native-fbsdk-next'
+
+import { GOOGLE_SIGNIN_WEBCLIENT_ID } from './Social.constants'
+
+export const facebookSignIn = async () => {
+  try {
+    const result = await LoginManager.logInWithPermissions(['email', 'public_profile'])
+    const currentProfile = await Profile.getCurrentProfile()
+    const accessToken = await AccessToken.getCurrentAccessToken()
+    // do nothing when cancelled
```
cameronolivier: we should log a failure action - even if we do nothing with it right now.

src/modules/Auth/Social/Social.config.ts
```diff
@@ -0,0 +1,40 @@
...
+    if (!result.isCancelled) {
+      const details = Object.assign({}, currentProfile, accessToken)
+      return details
+    }
+  } catch (error) {
+    throw new Error(error)
+  }
+}
+
+export const googleSignIn = async () => {
```
cameronolivier: likewise, this needs to be middleware - and can be in the auth middleware I think

src/modules/Auth/Social/Social.constants.ts
```diff
@@ -0,0 +1 @@
+export const GOOGLE_SIGNIN_WEBCLIENT_ID = '1007644634835-8b7vl8u89o7ah0d6pbhbtajarp23h3j0.apps.googleusercontent.com'
```
cameronolivier: I think we can pull all the Social folder into the Auth folder and file structure and flatten everything out.

src/modules/Auth/Social/Social.utils.test.ts
```diff
@@ -0,0 +1,110 @@
+import { GOOGLE_SIGNIN_WEBCLIENT_ID } from './Social.constants'
+import { Providers } from './Social.types'
+import * as SUT from './Social.utils'
+
+describe('modules/Social/Social.utils', () => {
+  describe('mapFacebookRegistrationData', () => {
```
cameronolivier: I'm not quite sure what this is doing? Should it not be called something like `selectRegistrationCredentialsFromFacebook` or similar? Although I can see how mapping applies here, I think 'select' is a better monicker, as it closely simulates what we're doing in our selectors from state, and have used in other places as well.

src/modules/Auth/Social/Social.utils.test.ts
```diff
@@ -0,0 +1,110 @@
...
+    it('should return the user registration credentials from facebook auth payload', () => {
+      const mockAuthData = {
```
cameronolivier: and please add some `given when then` for these with more details as to what's expected to happen.

src/modules/Auth/Social/Social.utils.test.ts
```diff
@@ -0,0 +1,110 @@
...
+        providerKey: 'DATA',
+        token: 'TOKEN',
+      }
+
+      const result = SUT.mapFacebookRegistrationData(mockAuthData)
+      expect(result).toEqual(mockMappedData)
+    })
+  })
+  describe('mapGoogleRegistrationData', () => {
+    it('should return the user registration credentials from google auth payload', () => {
```
cameronolivier: as above.

src/modules/Auth/Social/Social.utils.test.ts
```diff
@@ -0,0 +1,110 @@
...
+        provider: Providers.Google,
+        providerKey: GOOGLE_SIGNIN_WEBCLIENT_ID,
+        token: 'TOKEN',
+      }
+
+      const result = SUT.mapGoogleRegistrationData(mockAuthData)
+      expect(result).toEqual(mockMappedData)
+    })
+  })
+  describe('mapFacebookLoginData', () => {
```
cameronolivier: as above

src/modules/Auth/Social/Social.utils.test.ts
```diff
@@ -0,0 +1,110 @@
...
+        provider: Providers.Facebook,
+        providerKey: 'DATA',
+        token: 'TOKEN',
+      }
+
+      const result = SUT.mapFacebookLoginData(mockAuthData)
+      expect(result).toEqual(mockMappedData)
+    })
+  })
+  describe('mapGoogleLoginData', () => {
```
cameronolivier: as above.

src/modules/Auth/Social/Social.utils.test.ts
```diff
@@ -0,0 +1,110 @@
+import { GOOGLE_SIGNIN_WEBCLIENT_ID } from './Social.constants'
+import { Providers } from './Social.types'
+import * as SUT from './Social.utils'
+
+describe('modules/Social/Social.utils', () => {
+  describe('mapFacebookRegistrationData', () => {
+    it('should return the user registration credentials from facebook auth payload', () => {
+      const mockAuthData = {
+        firstName: 'DATA',
+        lastName: 'DATA',
```
cameronolivier: when you have multiple fields with 'DATA' values, it's impossible to know what your expected data is referring to or how it works.

cameronolivier: Rather use specific values (like `firstName: 'FIRST NAME'`) so it's obvious how you're deriving data in the expected.

src/modules/Auth/Social/Social.utils.test.ts
```diff
@@ -0,0 +1,110 @@
...
+        email: 'DATA',
+        imageURL: 'DATA',
+        applicationID: 'DATA',
+        middleName: 'DATA',
+        name: 'DATA',
+        userID: 'DATA',
+        accessToken: 'TOKEN',
+      }
+
+      const mockMappedData = {
```
cameronolivier: also best is to define this directly in your expected.

```js
expect(result).toEqual({        
        firstName: 'DATA',
        lastName: 'DATA',
        email: 'DATA',
        provider: Providers.Facebook,
        providerKey: 'DATA',
        token: 'TOKEN',
      })
```

Second prize is to call this 'expected' so that it's obvious what you're expecting out of the function.

cameronolivier: I'm sorry I'm being a stickler on the tests - but this is how we get to understand what our code is meant to be doing - if this isn't written properly the codebase becomes harder and harder to reason about.

src/modules/Auth/Social/Social.utils.ts
```diff
@@ -0,0 +1,54 @@
+import { curry, lensPath, map, view } from 'ramda'
+
+import { GOOGLE_SIGNIN_WEBCLIENT_ID } from './Social.constants'
+import { Providers } from './Social.types'
+
+const remap = curry((desc: any, obj: any) =>
+  map((path: any) => (typeof path === 'string' ? path : view(lensPath(path), obj)), desc),
+)
+export const mapFacebookRegistrationData = remap({
```
cameronolivier: it might be simpler to use ramda's `applySpec`, which takes an object and returns a new object with that data,and i already created, but up to you.

so it would be something like this:
```js
const selectGoogleRegistrationCredentials = applySpec({
  provider: always(Providers.Google),
  email: path(['user', 'email']),
  firstName: path(['user', 'givenName']),
  lastName: path(['user', 'familyName']),
  providerKey: always(GOOGLE_SIGNIN_WEBCLIENT_ID),
  token: prop('idToken'),
})
```

src/modules/Auth/Social/Social.utils.ts
```diff
@@ -0,0 +1,54 @@
+import { curry, lensPath, map, view } from 'ramda'
+
+import { GOOGLE_SIGNIN_WEBCLIENT_ID } from './Social.constants'
+import { Providers } from './Social.types'
+
+const remap = curry((desc: any, obj: any) =>
```
cameronolivier: if you're keeping this, it needs to be tested.

src/modules/Auth/Social/index.ts
```diff
@@ -0,0 +1,11 @@
+import { facebookSignIn, googleSignIn } from './Social.config'
+import { Providers } from './Social.types'
+
+export default (provider: string) => {
```
cameronolivier: I think this will be unnecessary when we refactor the below to use middleware. 

We should be firing off actions that handle side-effects. All app side-effects need to be handled within middleware.

cameronolivier: Unless specifc usecases where it's not applicable (like redux-persist, although I'm pretty sure that's using redux under the hood)

src/modules/Login/Login.tsx
```diff
@@ -9,17 +9,19 @@ import { ScrollView, View } from 'react-native'
...
 
 import Text, { BodyLevels, HeaderLevels, Link, TextAlign } from '../../components/Typography'
-import { AuthCredentials } from '../Auth/Auth.types'
+import { AuthCredentials, AuthSocialRegistrationCredentials } from '../Auth/Auth.types'
 import styles from './Login.styles'
 import LoginForm from './LoginForm/LoginForm'
 
 interface Props {
   navigation: StackNavigationProp<AuthNavigatorParamsList, AuthNavigationRoutes.Login>
+  onLoginWithSocial: (details: AuthSocialRegistrationCredentials) => void
```
cameronolivier: Why is this 'RegistrationCredentials' for login? Is that correct (if so, maybe we should rename it to `AuthSocialCredentials` if they are inter-changeable between login and reg

src/modules/Register/Register.tsx
```diff
@@ -44,7 +45,7 @@ const Register = ({ navigation }: Props) => {
         <Text.Body level={BodyLevels.small} style={styles.bodyText}>
           {t('registerSocial')}
         </Text.Body>
-        <SocialRegistration />
+        <SocialRegistration onRegisterWithSocial={onRegisterWithSocial} />
```
cameronolivier: This component could probably be called `SocialAuth` or `SSOButtons` or something similar (if you create a single component as suggested above)

src/redux/middleware.ts
```diff
@@ -18,6 +18,10 @@ const commonMiddleware: Middleware[] = [
 
 const featureModuleMiddleware = [
   authMiddleware.authLoginFlow,
+  authMiddleware.authSocialRegistrationSuccessFlow,
+  authMiddleware.authSocialLoginSuccessFlow,
+  authMiddleware.authSocialLoginFlow({ socialAuth: socialAuth, notification: showSimpleMessage }),
```
cameronolivier: Optional, but `socialAuth: socialAuth,` can just be `socialAuth,`

cameronolivier (changes requested):

src/components/SocialRegistration/SocialRegistration.tsx
sciphergfx: We can refactor this much later

src/components/SocialSSO/SocialSSO.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react'
```
cameronolivier: I think this is a little unnecessarily named -what I mean is it means Social Social Sign On, right?
So, I think SSO is actually good, or maybe SSOAuth?

src/modules/Auth/Auth.middleware.test.ts
```diff
@@ -282,6 +421,42 @@ describe('modules/Auth/Auth.middleware', () => {
       // when ... we respond to the authRegistration action
       await invoke(action)
 
+      // then ... the notification should be called
+      expect(mockNotification).toHaveBeenCalled()
+    })
+  })
+  describe('authSocialLoginFailureFlow', () => {
+    it('should correctly send a notification to the user', async () => {
```
cameronolivier: these shouldn't need to be async / await (please keep an eye on this in general)

src/modules/Auth/Social/Social.config.ts
```diff
@@ -0,0 +1,40 @@
+import { GoogleSignin, statusCodes } from '@react-native-google-signin/google-signin'
+import { AccessToken, LoginManager, Profile } from 'react-native-fbsdk-next'
+
+import { GOOGLE_SIGNIN_WEBCLIENT_ID } from './Social.constants'
+
+export const facebookSignIn = async () => {
```
cameronolivier: Why isn't this in the middleware - I think that's the best place for it?  These side-effects should be handled within a middleware - so that at least we can catch fails with failure actions

src/redux/store.ts
```diff
@@ -18,7 +19,17 @@ const store = configureStore({
...
     getDefaultMiddleware({
       serializableCheck: {
-        ignoredActions: [FLUSH, PAUSE, PERSIST, REGISTER, apiActions.apiRequest.type],
+        ignoredActions: [
+          FLUSH,
+          PAUSE,
+          PERSIST,
+          REGISTER,
+          apiActions.apiRequest.type,
+          authActions.authLoginSuccess.type,
```
cameronolivier: why are we ignoring these new actions?

cameronolivier (changes requested):

src/components/SocialSSO/SocialSSO.tsx
sciphergfx: noted

src/modules/Auth/Social/Social.config.ts
sciphergfx: the function is called in the middleware @cameronolivier

src/redux/store.ts
sciphergfx: Feedback from SSO auth has unserialised data

src/components/SSO/index.ts
```diff
@@ -0,0 +1,3 @@
+import SocialRegistration from './SSO'
```
cameronolivier: rename to `import SSO from './SSO'`

src/components/index.ts
```diff
@@ -23,9 +23,9 @@ import OnboardingForms from './OnboardingForms'
 import Optional from './Optional'
 import ProfilePhoto from './ProfilePhoto'
 import ProgressCircle from './ProgressCircle'
+import SocialRegistration from './SSO'
```
cameronolivier: We need to stop using this index file to export-  It's better to import directly from the component module as this creates circular references. This is a legacy thing that I think we need to move away from.

cameronolivier (changes requested):

cameronolivier (changes requested):
I've tested this locally - and both Google and Facebook registrations are failing - is this an issue with the API? 

I see it seems like Google SSO reg works... as in it gets the data back from google, but fails on submitting to the API so there's something that's not working here.

src/components/index.ts
sciphergfx: Noted, Do we create a separate branch to correct this across, or you want me to effect it on this branch?

src/components/index.ts
```diff
@@ -23,9 +23,8 @@ import OnboardingForms from './OnboardingForms'
 import Optional from './Optional'
 import ProfilePhoto from './ProfilePhoto'
 import ProgressCircle from './ProgressCircle'
+import SSO from './SSO'
```
cameronolivier: please don't export from here - rather export from the component module itself. This can cause super annoying circular dependency issues.

src/modules/Auth/Auth.middleware.test.ts
```diff
@@ -72,8 +78,147 @@ describe('modules/Auth/Auth.middleware', () => {
...
+      const create = createMiddlewareMock(jest)
+      const action = authSocialLoginSuccess(userSocialLoginData)
+      // @ts-ignore
+      const { invoke, store } = create(SUT.authSocialLoginSuccessFlow)
+      invoke(action)
+
+      expect(store.dispatch).toHaveBeenCalledWith(
+        ApiActions.apiRequest(
+          mergeRight(ApiAuthConstants.LOGIN_SOCIAL_CONFIG, {
+            isTokenRequired: false,
```
cameronolivier: You can remove this - all of the auth configs are preset to have `isTokenRequired: false,`:
```js
// from api/auth/auth.constants.ts

//This is the base Auth config object, merged into every endpoint config:
export const AUTH_CONFIG: Partial<ApiMeta> = {
  client: ApiClients.Auth,
  method: ApiMethods.Post,
  isTokenRequired: false,
}
```

src/modules/Auth/Auth.middleware.test.ts
```diff
@@ -72,8 +78,147 @@ describe('modules/Auth/Auth.middleware', () => {
       )
     })
   })
+  describe('authSocialLoginFlow', () => {
+    it('should correctly handle being called', () => {
+      // given ... the action is fired
```
cameronolivier, line 84: these tests should be broken into the following 
given - the state we expect the app to be in when we're testing (this includes redux state, etc)
when - this is the action - so this is what we want to be testing
then -  this is what we expect to happen _when_ the thing we're testing happens.

given when then should be in user-readable format - ie Clayton should be able to read it. Clayton should even be able to write the given/when/then criteria - which we fulfill in code. so we should avoid using language that simply reflects _what_ the code is _doing_ and rather reflect what the user/stakeholder needs from it. 

Please chat to me tomorrow morning if you're not quite sure how this works and we can walk through it.

We should strive to have all our tests follow this flow.

src/modules/Auth/SSOAuth/index.ts
```diff
@@ -0,0 +1,11 @@
+import { facebookSignIn, googleSignIn } from './SSOAuth.config'
+import { Providers } from './SSOAuth.types'
+
+export default (provider: string) => {
```
cameronolivier: OPTIONAL, but also I really don't want switches in our code if we can help it - a better way to write this is with a simple map function, so we avoid switch altogether:
```js
const ssoProviders = {
  facebook: () => facebookSignIn(),
  goodle: () => googleSignIn(),
}
```

Then in usage you just say 
```js
 ssoProviders[provider]()
```

src/modules/Auth/SSOAuth/SSOAuth.utils.ts
```diff
@@ -0,0 +1,51 @@
+import { always, applySpec, path, prop } from 'ramda'
+
+import { GOOGLE_SIGNIN_WEBCLIENT_ID } from './SSOAuth.constants'
+import { Providers } from './SSOAuth.types'
+
+export const selectRegistrationCredentialsFromFacebook = applySpec({
```
cameronolivier: should be `extract...`  - and likewise below

src/modules/Auth/SSOAuth/SSOAuth.utils.ts
```diff
@@ -0,0 +1,51 @@
...
+  token: prop('accessToken'),
+})
+
+export const selectLoginCredentialsFromGoogle = applySpec({
+  provider: always(Providers.Google),
+  providerKey: always(GOOGLE_SIGNIN_WEBCLIENT_ID),
+  token: prop('idToken'),
+})
+
+export const selectSocialLoginCredentials = (authProvider: Providers, authData: any) => {
```
cameronolivier: same as comment on switches below - if you export this as an object, the 'provider' becomes the key, and you can simply call it as :
```js
selectSocialLoginCredentials[provider](authData)
```

src/modules/Auth/SSOAuth/SSOAuth.config.ts
```diff
@@ -0,0 +1,40 @@
+import { GoogleSignin, statusCodes } from '@react-native-google-signin/google-signin'
+import { AccessToken, LoginManager, Profile } from 'react-native-fbsdk-next'
+
+import { GOOGLE_SIGNIN_WEBCLIENT_ID } from './SSOAuth.constants'
+
+export const facebookSignIn = async () => {
```
cameronolivier: I keep coming back to it - this needs to be in middleware. this is not the place to be doing this. 

Please refactor. 

I get there may seem to be some similarities with the api.config - but we're not actually calling any async/ data functions in the config- I'm just front-loading/setting up the api instance that is then only called within the middleware. 

The way you've set this up is an antipattern. 

Sorry to circle back to this so late in the game, but every time I come back to this - it just doesn't sit right with the rest of the architecture of the app - and I suspect it might be some of the reason we're having issues in the app flow.

cameronolivier (changes requested):

src/modules/Auth/Auth.middleware.test.ts
cameronolivier: that said ^^ I'm happy with flexibility as we move towards this everywhere - let's aim at getting better with this - at the moment the test quality is a bit all over the place - some of the tests are spot on, others aren't really structured well - so I'll give a bit of pushback, but I'm not going to stop it going through for now. Just please focus on this getting better. and I'll keep giving feedback where it could be better, but as long as we're testing the right things - and what we're testing covers the functionality, I'm going to stop requesting changes to tests as something that stops merging (for now) - I think it's something we can work towards.

src/modules/Auth/SSOAuth/SSOAuth.config.ts
sciphergfx: I disagree the way it's wired up is an anti-pattern, happy to chat about it.

src/modules/Auth/SSOAuth/index.ts
sciphergfx: This is valid, similarly having to just call 
```js
 ssoProviders(providerName)
```
looks cleaner

src/redux/store.ts
cameronolivier: Can we not extract the functions from the payload before dispatching it in the action? I'd rather sanitize the data if possible than add actions here unnecessarily.

src/components/index.ts
cameronolivier: I think we should refactor as we go, for now. Correct it when we get to it and don't add new, but we should do the big refactor in a separate branch, with a jira ticket, definitely.

src/modules/Auth/Auth.middleware.ts
```diff
@@ -48,6 +56,118 @@ export const authLoginFlow: Middleware =
     }
     return result
   }
+
+export const authSocialLoginFlow =
+  ({ SSOAuth }: { SSOAuth: Function; notification: typeof showSimpleMessage }): Middleware =>
```
cameronolivier: we shouldn't be uppercasing utils/dependencies functions (as you know :) ) - I think we both just missed this.

src/modules/Auth/SSOAuth/SSOAuth.constants.ts
```diff
@@ -0,0 +1,6 @@
+export const GOOGLE_SIGNIN_WEBCLIENT_ID = '1007644634835-8b7vl8u89o7ah0d6pbhbtajarp23h3j0.apps.googleusercontent.com'
```
cameronolivier: is this a public-facing id or something we need to save securely? 
If it's public-facing it should be in our ENV config.
if it needs to be saved securely, then we need to set it in the Circle CI config (and in our 1pass details)

src/vendor.d.ts
```diff
@@ -1,11 +1,13 @@
 declare module 'ramda' {
   export function __(...rest: any[]): any[]
   export function always(...rest: any[]): any[]
+  export function applySpec(...rest: any[]): any[]
```
cameronolivier: remove the `[]` off the end:
```js
export function applySpec(...rest: any[]): any
```

Will fix the linting errors in the utils.

src/modules/Auth/SSOAuth/SSOAuth.utils.ts
```diff
@@ -0,0 +1,89 @@
+import { always, applySpec, merge, path, prop } from 'ramda'
+
+import { facebookAuthDependencies, googleAuthDependencies } from './../../SSOAuth/SSOAuth.types'
```
cameronolivier: fix this import, please

cameronolivier: and types should be `PascalCase`

cameronolivier (changes requested):
Currently when we cancel, it throws an error:
Screenshot 2021-07-12 at 22 53 01

We shouldn't be throwing an error on a cancelled login/registration.

Also, I cancelled facebook, but the success action was fired:
Screenshot 2021-07-12 at 23 16 55
The same happens with social logins. 
We should probably log a cancelled action - and ignore it. (or simply ignore it) 

Then I tried to register with Facebook.- the SSO part went fine but our API returned with this error:
```js
success: false
code: 400
message: "The LastName field is required., The FirstName field is required."
```
So it doesn't look like the correct information is being submitted.

Then have found some further issues in the code. Mostly naming convention stuff. As there are still issues to iron out, please do them in this PR as well.

And please run through your PRs and make sure that these things are correct before you request a review.

src/api/api.utils.ts
```diff
@@ -1,26 +1,28 @@
...
-  equals,
-  mergeAll,
 } from 'ramda'
 
 import { RootState } from '../redux/redux.types'
 import { StdObj } from '../types/general.types'
-import { ApiClientArgs, ApiMeta, ApiCall, PrepareApiRequestData } from './api.types'
+import { ApiCall, ApiClientArgs, ApiMeta, GenerateEndpoint, PrepareApiRequestData } from './api.types'
+
+export const generateEndpoint: GenerateEndpoint = join('/')
```
cameronolivier: I think this was a merge issue - this util can be removed - it is no longer used - I removed it in a previous PR.

src/api/api.utils.ts
```diff
@@ -1,26 +1,28 @@
...
-  always,
-  ifElse,
-  equals,
-  mergeAll,
 } from 'ramda'
 
 import { RootState } from '../redux/redux.types'
 import { StdObj } from '../types/general.types'
-import { ApiClientArgs, ApiMeta, ApiCall, PrepareApiRequestData } from './api.types'
+import { ApiCall, ApiClientArgs, ApiMeta, GenerateEndpoint, PrepareApiRequestData } from './api.types'
```
cameronolivier: You can remove the GenerateEndpoint as well - as it actually no longer exists.

src/api/api.utils.ts
cameronolivier: I presume that you didn't intend to make any changes to this file - so could roll back to whatever state develop is in for this file - and that should be fine.

src/modules/Auth/Auth.middleware.test.ts
```diff
@@ -1,5 +1,5 @@
 import { resetAppData } from 'modules/App/App.reducer'
-import { INITIAL_STATE } from 'modules/Auth/Auth.reducer'
+import { authSocialLoginFailure, authSocialRegistrationFailure, INITIAL_STATE } from 'modules/Auth/Auth.reducer'
```
cameronolivier, line 2: please fix the imports here - this is the wrong place to be importing from - also we are already importing from the reducer file, below - so merge in with that, please.

src/modules/Auth/Auth.middleware.test.ts
```diff
@@ -277,14 +419,14 @@ describe('modules/Auth/Auth.middleware', () => {
       // @ts-ignore
       const { invoke } = create(SUT.authRegistrationSuccessFlow({ notification: mockNotification }))
 
-      await invoke(action)
+      invoke(action)
 
-      // then ... the notification should be called
+      // then ... the notification should be calleds
```
cameronolivier: remove the rogue s

src/modules/Auth/SSOAuth/SSOAuth.utils.ts
```diff
@@ -0,0 +1,89 @@
...
+}
+
+export const onFacebookAuth = async ({ fbLoginManager, fbProfile, fbAccessToken }: facebookAuthDependencies) => {
+  try {
+    const loginResponse = await fbLoginManager.logInWithPermissions(FACEBOOK_PERMISSIONS)
+
+    const userProfile = await fbProfile.getCurrentProfile()
+    const accessToken = await fbAccessToken.getCurrentAccessToken()
+
+    if (!loginResponse.isCancelled) {
```
cameronolivier: I've mentioned this elsewhere, but we aren't handling if this is cancelled in anyway (and so our middleware fires success). It should simply revert quietly.

cameronolivier (changes requested):

src/modules/Auth/SSOAuth/SSOAuth.constants.ts
sciphergfx: I'm not sure if it needs to be saved securely, but we can do that. It acts more like a url would @cameronolivier

src/modules/Auth/Auth.middleware.test.ts
cameronolivier: This still needs to be done. Add it with the below import from the Auth reducer

src/modules/SSOAuth/SSOAuth.types.ts
```diff
@@ -0,0 +1,17 @@
+import { GoogleSignin } from '@react-native-google-signin/google-signin'
+import { AccessToken, LoginManager, Profile } from 'react-native-fbsdk-next'
+
+export enum Providers {
+  Google = 'google',
+  Facebook = 'facebook',
+}
+
+export type FacebookAuthDependencies = {
+  fbLoginManager: typeof LoginManager | any
```
cameronolivier: if you or as `any` then you may as well just make it `any`.

src/modules/SSOAuth/SSOAuth.utils.test.ts
```diff
@@ -0,0 +1,182 @@
...
+          configure: jest.fn(),
+          hasPlayServices: jest.fn(),
+          signIn: jest.fn().mockResolvedValue(authResponse),
+        },
+      }
+      const result = await SUT.onGoogleAuth(mockGoogleConfigStub)
+      //then ... validate user authentication data is returned
+      expect(result).toEqual(authResponse)
+    })
+    it('should correctly handle cancelled google auth', async () => {
```
cameronolivier, line 180: we don't want to throw an error if the user cancels. it's not an error. it's just a canceled event. So if you want to dispatch a canceled action so we have access to that information, then I think that's a good idea - but otherwise it should just fail silently. We shouldn't fire a success or an error - because it is neither.

src/modules/SSOAuth/SSOAuth.utils.ts
```diff
@@ -0,0 +1,85 @@
...
+    const loginResponse = await fbLoginManager.logInWithPermissions(FACEBOOK_PERMISSIONS)
+
+    const userProfile = await fbProfile.getCurrentProfile()
+    const accessToken = await fbAccessToken.getCurrentAccessToken()
+
+    if (!loginResponse.isCancelled) {
+      const authResponse = merge(userProfile, accessToken)
+      return authResponse as object
+    } else {
+      throw new Error(fbAuthCancelledErrorMessage)
```
cameronolivier: as mentioned above - don't throw an error on cancel. return false maybe and handle that or ignore that in the middleware - but a user cancelling is not an error event.

src/modules/SSOAuth/SSOAuth.utils.ts
```diff
@@ -0,0 +1,85 @@
...
+    }
+  } catch (error) {
+    throw new Error(error.message)
+  }
+}
+
+export const onGoogleAuth = async ({ googleSignIn }: GoogleAuthDependencies) => {
```
cameronolivier: how are we handling a cancelled event with google?

cameronolivier (changes requested):
The primary thing here is: we should not be throwing an error if the user cancels. It should fail silently. We can dispatch a cancelled action if we want to deal with it down the line, but not an error.

src/modules/SSOAuth/SSOAuth.utils.test.ts
```diff
@@ -0,0 +1,184 @@
...
+      const result = SUT.extractLoginCredentialsFromGoogle(mockAuthData)
+      //then result should equal facebook login data
+      expect(result).toEqual({
+        provider: Providers.Google,
+        providerKey: GOOGLE_SIGNIN_WEBCLIENT_ID,
+        token: 'TOKEN',
+      })
+    })
+  })
+  describe('onFacebookAuth', () => {
```
cameronolivier, line 112: we need to add a test for if it errors out (that if any of our dependencies throw an error, we're catching it as expected)

src/modules/SSOAuth/SSOAuth.utils.test.ts
```diff
@@ -0,0 +1,184 @@
...
+        fbProfile: { getCurrentProfile: jest.fn().mockResolvedValue({ profile: 'USER_PROFILE' }) },
+        fbAccessToken: { getCurrentAccessToken: jest.fn().mockResolvedValue({ token: 'FACEBOOK_TOKEN' }) },
+      }
+
+      const result = await SUT.onFacebookAuth(mockFacebookConfigStub)
+      //then return false
+      expect(result).toBe(false)
+    })
+  })
+  describe('onGoogleAuth', () => {
```
cameronolivier, line 156: as with facebook we also need to test that we are throwing an error when we'd expect to throw one.

cameronolivier (changes requested):
just need to test the failing states of the sso auth utils

src/modules/Auth/Auth.middleware.test.ts
cameronolivier: @sciphergfx this needs to be fixed. But let's refactor when we next touch this file.

src/modules/SSOAuth/index.ts
```diff
@@ -0,0 +1,14 @@
+import { GoogleSignin, statusCodes } from '@react-native-google-signin/google-signin'
+import { AccessToken, LoginManager, Profile } from 'react-native-fbsdk-next'
+
+import { Providers } from './SSOAuth.types'
+import { onFacebookAuth, onGoogleAuth } from './SSOAuth.utils'
+
+export default (provider: string) => {
```
cameronolivier, line 7: crap I missed this.
we should define this function, and it should be tested. 

But let's leave it - and we can update when we refactor SSO

cameronolivier (dismissed):

cameronolivier (dismissed):

src/modules/Register/Register.tsx
```diff
@@ -1,15 +1,15 @@
...
 import Button from 'components/Button'
+import LargeHeader from 'components/LargeHeader'
 import SSO from 'components/SSO'
 import { AuthNavigationRoutes, AuthNavigatorParamsList } from 'modules/AuthNavigation/AuthNavigation.types'
 import React from 'react'
 import { useTranslation } from 'react-i18next'
 import { View } from 'react-native'
 import { Colors } from 'styles'
 
-import LargeHeader from '../../components/LargeHeader'
```
cameronolivier, line 11: this is the preferred way to import. 
but happy for now.

cameronolivier (approved):

PR #73, opened 2021-06-28
YOMA-436: refactor(User): user data in profile view

sciphergfx:
## Work Done
- created user module
- tested middleware and other utilities
- Plugged user data to profile view


src/modules/Profile/Profile.container.tsx
```diff
@@ -0,0 +1,14 @@
+import { selectUserState } from 'modules/User/User.selector'
+import { connect } from 'react-redux'
+import { RootState } from 'redux/redux.types'
+
+import Profile from './Profile'
+
+const mapStateToProps = (state: RootState) => {
```
cameronolivier: as with others, you should create a Profile.selector that handles mapping data from state to props.

cameronolivier (changes requested):
just a smallish thing

src/modules/User/User.middleware.test.ts
```diff
@@ -0,0 +1,41 @@
...
+      const create = createMiddlewareMock(jest)
+      const credentials = defaultUserLoginResponseData
+      const action = authLoginSuccess(credentials)
+      // @ts-ignore
+      const { invoke, next } = create(SUT.setUserOnAuthFlow)
+
+      // when ... we respond to the authLoginSuccess action
+      await invoke(action)
+
+      // then ...validate setUserOnAuthFlow
```
cameronolivier, line 22: we should also check that store.dispatch was called. - just check it was called, that's all we need to do here.

src/modules/User/User.middleware.ts
```diff
@@ -0,0 +1,17 @@
+import { authLoginSuccess } from 'modules/Auth/Auth.reducer'
+import { Middleware } from 'redux'
+
+import { setUserCredentials } from './User.reducer'
+import { selectUserCredentialsFromLoginPayload } from './User.utils'
+
+export const setUserOnAuthFlow: Middleware =
+  ({ dispatch }) =>
+  next =>
+  async action => {
```
cameronolivier: no need for async

src/modules/User/User.middleware.test.ts
```diff
@@ -0,0 +1,41 @@
...
+      // given ... the authLogin action is fired
+      const create = createMiddlewareMock(jest)
+      const credentials = defaultUserLoginResponseData
+      const action = authLoginSuccess(credentials)
+      // @ts-ignore
+      const { invoke, store } = create(SUT.setUserOnAuthFlow)
+      const userData = selectUserCredentialsFromLoginPayload(action)
+
+      // when ... we respond to the authLoginSuccess action
+      await invoke(action)
```
cameronolivier: can remove async await

cameronolivier (changes requested):
a couple small changes - but otherwise good

src/modules/User/User.reducer.ts
```diff
@@ -0,0 +1,14 @@
+import { createAction, createReducer } from '@reduxjs/toolkit'
+import { UserResponse } from 'modules/Auth/Auth.types'
+import { mergeDeepRight } from 'ramda'
+
+const name = '[User]'
+export const INITIAL_STATE = {}
+
+export const setUserCredentials = createAction<UserResponse>(`${name} Set User Credentials`)
```
cameronolivier: std naming convention for `createAction()` is to be: `[<name of module>] <name of action>`

eg: 
```js
export const setUserCredentials = createAction<UserResponse>(`${name} setUserCredentials`)
```

This keeps it simple and a 1:1 reference in the logger to the js action name.

cameronolivier (changes requested):

src/modules/User/User.middleware.test.ts
cameronolivier: this hasn't been done?

src/modules/User/User.middleware.ts
```diff
@@ -0,0 +1,17 @@
...
+import { selectUserCredentialsFromLoginPayload } from './User.utils'
+
+export const setUserOnAuthFlow: Middleware =
+  ({ dispatch }) =>
+  next =>
+  action => {
+    const result = next(action)
+    if (authLoginSuccess.match(action)) {
+      const credentials = selectUserCredentialsFromLoginPayload(action)
+      dispatch(setUserCredentials(credentials))
```
cameronolivier: And this should be `setUser`.

src/modules/User/User.middleware.ts
```diff
@@ -0,0 +1,17 @@
...
+import { setUserCredentials } from './User.reducer'
+import { selectUserCredentialsFromLoginPayload } from './User.utils'
+
+export const setUserOnAuthFlow: Middleware =
+  ({ dispatch }) =>
+  next =>
+  action => {
+    const result = next(action)
+    if (authLoginSuccess.match(action)) {
+      const credentials = selectUserCredentialsFromLoginPayload(action)
```
cameronolivier: this probably isn't 'credentials' but rather user data = simply 'user' - credentials are username/password.

src/modules/User/User.middleware.ts
```diff
@@ -0,0 +1,17 @@
+import { authLoginSuccess } from 'modules/Auth/Auth.reducer'
+import { Middleware } from 'redux'
+
+import { setUserCredentials } from './User.reducer'
+import { selectUserCredentialsFromLoginPayload } from './User.utils'
```
cameronolivier: as below - remove 'credentials' as we're getting user data from auth success, not credentials, correct?

src/modules/User/User.reducer.test.ts
```diff
@@ -0,0 +1,20 @@
+import { USER_RESPONSE } from 'modules/Profile/Profile.constants'
+
+import SUT, { setUserCredentials } from './User.reducer'
+
+describe('modules/User/User.reducer', () => {
+  describe('setUserCredentials', () => {
```
cameronolivier: follow action renaming here as well (`setUser`)

src/modules/User/User.selector.ts
```diff
@@ -0,0 +1,3 @@
+import { RootState } from 'redux/redux.types'
+
+export const selectUserState = (state: RootState) => state.user
```
cameronolivier: this should just be `selectUser` - state is superfluous here - and creates unnecessary ambiguity for if we ever have a 'user state' (ie subscribed, or onboarded, or what-have-you) - that said, in that event, I think user status might be better- but the Api might dictate.

src/modules/User/User.utils.ts
```diff
@@ -0,0 +1,3 @@
+import { path, pipe } from 'ramda'
+
+export const selectUserCredentialsFromLoginPayload = pipe(path(['payload', 'data', 'data', 'user']))
```
cameronolivier: as mentioned above - this should be named correctly to what it's doing - 'credentials' is incorrect so should be removed. 

```js
selectUserFromLoginPayload
```

src/modules/User/User.reducer.ts
```diff
@@ -0,0 +1,14 @@
+import { createAction, createReducer } from '@reduxjs/toolkit'
+import { UserResponse } from 'modules/Auth/Auth.types'
+import { mergeDeepRight } from 'ramda'
+
+const name = '[User]'
+export const INITIAL_STATE = {}
```
cameronolivier: I think we should create an initial state that reflects the user state - but with empty data. 
So it needs to reflect the properties we expect the state to contain, with useful default values:
```js
id:"",
firstName:"",
lastName:"",
phoneNumber:null,
biography:null,
countryAlpha2:"",
email:"",
zltoWalletId:null,
zltoBalance:0,
covidChallengeCertificateURL:null,
tideChallengeCertificateURL:null,
photoURL:null,
organisationId:null,
organisation:null,
organisationVerified:false,
createdAt:"", // (or null)
lastLogin:"", // (or null)
```

src/modules/Profile/Profile.selector.tsx
```diff
@@ -0,0 +1,6 @@
+import { createSelector } from '@reduxjs/toolkit'
+import * as UserSelectors from 'modules/User/User.selector'
+
+export default createSelector(UserSelectors.selectUserState, user => ({
+  user,
```
cameronolivier: do we need the entire user object? We should only be bringing in those values that we will be needing in this view. 
Individually, or in a user object. So as individual props:
```js
export default createSelector( ... firstName, lastName, number... => ({
 firstName,
 lastName,
 number,
 ...
})
```
or in a user object: 
```js
export default createSelector( ... firstName, lastName, number... => ({
 user: {
   firstName,
   lastName,
   number,
   ...
 },
 ...
})
```

cameronolivier (changes requested):

src/modules/User/User.selector.ts
cameronolivier: It should also be tested. 

All selectors need to have tests.

cameronolivier (changes requested):

src/modules/Profile/Profile.selector.tsx
sciphergfx: I'm not sure if this matters, as it creates a reference for all user (updated) credentials. We can equally use other props else where in the application @cameronolivier

src/modules/User/User.middleware.test.ts
sciphergfx: I missed this one

src/modules/User/User.selector.ts
sciphergfx: I purposely don't create tests for rootState selector, as it'll always return same prop as defined in the reducer, but happy to add one @cameronolivier

src/modules/User/User.reducer.ts
sciphergfx: If we take this approach then we should initiate other reducer state values across, Your thoughts? @cameronolivier

src/modules/Profile/Profile.selector.tsx
sciphergfx: I'll implement this on YOMA-437

cameronolivier: Please do the changes here. Doing changes in future PRs makes it really difficult to track comments and changes across PRs.

src/modules/User/User.utils.ts
```diff
@@ -0,0 +1,3 @@
+import { path, pipe } from 'ramda'
+
+export const selectUserFromLoginPayload = pipe(path(['payload', 'data', 'data', 'user']))
```
cameronolivier: no need for a pipe here as it's just the 1 fn

src/modules/User/User.reducer.ts
cameronolivier: yes 100%. 

the only place I think we shouldn't (At least now) is the auth reducer purely because it's auth. But even there we could create the object - but yes, we should have our default INITIAL_STATE represent what the default values should be for each of our state properties

src/modules/User/User.reducer.test.ts
```diff
@@ -0,0 +1,19 @@
+import { USER_RESPONSE } from './../Profile/Profile.constants'
+import SUT, { setUser } from './User.reducer'
+
+describe('modules/User/User.reducer', () => {
+  describe('setUser', () => {
+    it('should return user data from login payload', () => {
+      // given ... there are are credentials in user state
+      const state = {}
```
cameronolivier, line 8: state is RootState - so should reflect the state, like:
```js
const state = {
  auth: AUTH_INITIAL_STATE,
  user: USER_INITIAL_STATE,
}
```

src/modules/User/User.middleware.ts
```diff
@@ -0,0 +1,17 @@
...
+import { setUser } from './User.reducer'
+import { selectUserFromLoginPayload } from './User.utils'
+
+export const setUserOnAuthFlow: Middleware =
+  ({ dispatch }) =>
+  next =>
+  action => {
+    const result = next(action)
+    if (authLoginSuccess.match(action)) {
+      const credentials = selectUserFromLoginPayload(action)
```
cameronolivier: small thing - but rename 'credentials' as well - to `user`.

cameronolivier (changes requested):

src/modules/User/User.reducer.test.ts
sciphergfx: In order to keep consistency, I think this change is required in many places, why don't we create a branch and effect them? I'll implement here nonetheless. I also see PR85 has rootStateFixture, We can use that here? @cameronolivier

cameronolivier (dismissed):

cameronolivier (approved):

PR #78, opened 2021-07-01
YOMA-449 - refactor(About) - About - should correctly display current data

sciphergfx:
## Work Done
Wired biography content to about widget


src/modules/About/About.container.tsx
```diff
@@ -0,0 +1,9 @@
+import selector from 'modules/About/About.selector'
```
cameronolivier: please import this locally `import selector from './About.selector'`

src/modules/About/About.selector.test.ts
```diff
@@ -0,0 +1,33 @@
+import * as SUT from './About.selector'
+
+describe('modules/About/About.selector', () => {
+  describe('selector ', () => {
+    it('should select props as expected when available in state', () => {
```
cameronolivier, line 6: Add a test for empty state (reflecting user: INITIAL_STATE)

cameronolivier (changes requested):

src/modules/About/About.container.tsx
sciphergfx: my IDE auto import seem to be causing this issue. I'll make changes, thanks

src/modules/About/About.selector.test.ts
sciphergfx: noted

src/modules/About/About.container.tsx
cameronolivier: I think it's a babel import thing that Ranjit added (that I'd rather remove) - but it's going to affect multiple files - so we need to keep an eye on it

cameronolivier (dismissed):

src/modules/User/User.selector.ts
```diff
@@ -1,3 +1,6 @@
+import { createSelector } from '@reduxjs/toolkit'
+import { propOr } from 'ramda'
 import { RootState } from 'redux/redux.types'
 
 export const selectUser = (state: RootState) => state.user
+export const selectBiography = createSelector(selectUser, propOr('', 'biography'))
```
cameronolivier, line 6: you don't need to use prop or if your initial reducer state is correct - because even if you don't have data, it'll always be an empty string - so what I'd suggest is set 'biography' to '' in the INITIAL_STATE object for the user.

Happy for you to do this in a follow up PR.

cameronolivier (approved):
Left a comment - please do this in #82

PR #82, opened 2021-07-01
YOMA-442 - refactor(DigitalCV): MyCV - Display real data in About widget

sciphergfx:
## Work Done
 - Display real data in About widget


cameronolivier (commented):
Happy with this, just want to get `YOMA-449` branch in and then I'll run it on device, so don't merge in  - let's merge to develop.

src/modules/DigitalCv/DigitalCv.container.tsx
```diff
@@ -0,0 +1,9 @@
+import selector from 'modules/About/About.selector'
```
cameronolivier: you're pulling this in from the wrong place - should be using the DigitalCV selector

src/modules/DigitalCv/DigitalCv.selector.ts
```diff
@@ -0,0 +1,7 @@
+import { createSelector } from '@reduxjs/toolkit'
+
+import * as userSelectors from '../User/User.selector'
```
cameronolivier: we should be importing this from the module index, not the internal selector.

cameronolivier (changes requested):
Picked up a couple small things.

src/modules/DigitalCv/DigitalCv.selector.ts
sciphergfx: Do you mean we should introduce selectors on the index.js?
```js
export { actions, reducer, middleware, selectors }
```

src/modules/About/About.selector.test.ts
```diff
@@ -46,7 +46,7 @@ describe('modules/About/About.selector', () => {
       // when ... we call the selector
       const result = SUT.default(state)
       // then ... should return result as expected
-      expect(result).toEqual({ biography: '' })
+      expect(result).toEqual({ biography: undefined })
```
cameronolivier: This tests when state (root state) is an entirely empty object - that would never be the case. state will always be an initial state object, or a populated state. 

So you should be wrapping selectors with `rootStateFixture()` and adding in data where you need it. 

So `ln 8` should be something like: 
```js
const stateMock = rootStateFixture({
        user: {
          id: 'USER_ID',
          firstName: 'FIRST_NAME',
          lastName: 'LAST_NAME',
          phoneNumber: 'PHONE_NUMBER',
          biography: 'BIOGRAPHY',
          countryAlpha2: 'COUNTRY_ALPHA2',
          email: 'USER_EMAIL@SOMEWHERE.TEST',
          zltoWalletId: 'ZLTO_WALLET_ID',
          zltoBalance: 1000,
          covidChallengeCertificateURL: 'COVID_CHALLENGE_CERTIFICATE_URL',
          tideChallengeCertificateURL: 'TIDE_CHALLENGE_CERTIFICATE_URL',
          photoURL: 'PHOTO_URL',
          role: 'ROLE',
          organisation: 'ORGANISATION',
          createdAt: 'CREATED_AT',
          lastLogin: 'LAST_LOGIN',
        },
      })
```

cameronolivier: And `ln 33`, you should be simply:
```js
const stateMock = rootStateFixture()
``` 
This will set up the state mock to be the equivalent of
```js
{
  about: ABOUT_INITIAL_STATE,
  user: USER_INITIAL_STATE,
}
```
which is what we'd expect an empty redux state to look like.

cameronolivier: Then we can remove this, because it doesn't test a state we could expect to have in the app. 
```js
it('should handle empty state if props not available', () => {
      // given ...
      const state = {}
      // when ... we call the selector
      const result = SUT.default(state)
      // then ... should return result as expected
      expect(result).toEqual({ biography: undefined })
    })
```

src/modules/User/User.selector.ts
```diff
@@ -1,6 +1,6 @@
 import { createSelector } from '@reduxjs/toolkit'
-import { propOr } from 'ramda'
+import { prop } from 'ramda'
 import { RootState } from 'redux/redux.types'
 
 export const selectUser = (state: RootState) => state.user
-export const selectBiography = createSelector(selectUser, propOr('', 'biography'))
+export const selectBiography = createSelector(selectUser, prop('biography'))
```
cameronolivier: we need to add a test for an empty user state here (well an `INITIAL_STATE`) to test that we're returning an empty string for the `biography` property.

cameronolivier: We don't have one yet.

cameronolivier (changes requested):
We need a couple of changes around tests.

src/modules/About/About.selector.test.ts
```diff
@@ -1,11 +1,11 @@
-import { INITIAL_STATE } from '../User/User.reducer'
+import { rootStateFixture } from './../../redux/redux.test.fixtures'
```
cameronolivier: we don't need this `./../....` in front of import strings. can simply go `../....`

cameronolivier: I'm happy to let it slide for now - but you need to be catching these things yourself before requesting a review.

cameronolivier (dismissed):

src/modules/DigitalCv/DigitalCv.tsx
```diff
@@ -1,19 +1,19 @@
 import { StackNavigationProp } from '@react-navigation/stack'
+import { CvCard, FirstTimeCard, HomeHeader, ViewContainer } from 'components'
+import { HomeNavigationRoutes, HomeNavigatorParamsList } from 'modules/HomeNavigation/HomeNavigation.types'
 import React from 'react'
 import { useTranslation } from 'react-i18next'
 import { ScrollView } from 'react-native'
 
-import { ViewContainer, HomeHeader, FirstTimeCard } from '../../components'
-import CvCard from '../../components/CvCard'
```
cameronolivier, line 7: This is the correct import - importing from 'components' breaks the app.

src/components/CvCard/CvCard.tsx
```diff
@@ -8,6 +8,7 @@ import Text, { TextAlign } from '../Typography'
 import styles from './CvCard.styles'
 
 type Props = {
+  text: string
```
cameronolivier: There are issues in implementing this card in the DigitalCV because this field is required. Initially the user won't have text, but more importantly - this breaks all the other CVCards in the DigitalCV view.

So options are - set a default text value of '' for all the cards in DigitalCV, or make this optional.  If we always expect there to be some text here (even if it might be '') then keep it as required. 

BTW - what are we going to do with the Education or MySkills sections where the content isn't text, but an array of presumably objects - to display the different credentials. I think that text might not be the correct property. 

I'd call the `text` field something like 'content' to cater for the list of info OR the string in the case of biography and the `defaultText` - `fallbackText` rather as 'defaultText' implies that we're expecting text and this is the default - fallback is frequently used for when there is no value (regardless of the data) - placeholderText could also work.

cameronolivier (changes requested):
I can't really test this because I don't have any data - so need the other task in first (for editing the about information) before I could confirm that this is working. Have you been able to test it @sciphergfx ?

src/modules/DigitalCv/DigitalCv.tsx
sciphergfx: Weird, sorted this too. I'll revert

src/components/CvCard/CvCard.tsx
```diff
@@ -16,7 +17,22 @@ type Props = {
   onEdit: () => void
 }
 
+const renderTextContent = (content: string, defaultText: string) => (
```
cameronolivier, line 20: happy with this for now. 

Honestly the easiest I think would have been to pass the content as children, and show a fallback if the children were null/empty. I know I mentioned content - I realise that's wrong. The designs are different for each credential - so it's probably best to pass in as children and we can handle the list renders in separate components that cater to each of the specific layouts. 
Screenshot 2021-07-13 at 21 22 29


THAT said, let's get back to this when we get to those components. For now, as long as the about text is being correctly rendered in the CvCard, I'm happy.

src/components/Typography/Body/Body.tsx
```diff
@@ -3,11 +3,13 @@ import { TextStyle } from 'react-native'
 
 import { GetComponentProps, WithChildren } from '../../../types/react.types'
 import Text from '../Text'
+import { TextAlign } from '../Text.types'
 import styles from './Body.styles'
 import { BodyLevels } from './Body.types'
 
 type Props = WithChildren<
   GetComponentProps<typeof Text> & {
+    align?: TextAlign
```
cameronolivier, line 12: there is no need to add this as `align` is already exposed on the Text component in `Typography/Text.tsx` and all child components (Body, Header, etc) accept all the same props as the base text component.

cameronolivier: Please revert these changes.

cameronolivier: This works: 
```js
<Text.Body align={TextAlign.left}>Hello World</Text.Body>
```

cameronolivier (changes requested):
small changes - basically just need to revert the changes in Body.tsx text

src/components/Typography/Body/Body.tsx
sciphergfx: There's a lint error here

src/modules/DigitalCv/DigitalCv.selector.ts
cameronolivier: yes. Where possible - but I think where we end up with circle imports, then we can import directly from the selector file.

src/modules/User/User.selector.test.ts
```diff
@@ -1,32 +1,13 @@
 import { rootStateFixture } from 'redux/redux.test.fixtures'
 
-import { USER_RESPONSE } from './../Profile/Profile.constants'
 import * as SUT from './User.selector'
 import { USER_RESPONSE } from './User.test.fixtures'
 
 describe('modules/User/User.selector', () => {
-  describe('selectUser ', () => {
```
cameronolivier, line 7: Why are we deleting this test?

cameronolivier (commented):
Not sure why the selectUser test was deleted. We shouldn't be?

src/modules/User/User.selector.test.ts
sciphergfx: It appeared twice, it's down below. Can reposition it

cameronolivier: Ok. It should be up here - but we can do that another time.

cameronolivier (approved):

PR #85, opened 2021-07-02
YOMA-483 - chore: navigation and fixing root state and fixtures

cameronolivier:
## Scope [YOMA-483](https://yomaworld.atlassian.net/browse/YOMA-483)

In doing the other task, wanted to get some things out the way first.
- explicitly set redux state type
-  rename AppNavigation to Navigation (keeping the naming terse)
- moved AppNavigation/Authentication into modules/AuthNavigation  - as we should keep the module structure flat.
- added a state fixture that frontloads state with all INITIAL_STATE for the app redux endpoints - means we have a clean slate starting point that we can override with whatever state we're trying to test.

## Work Done
renaming AppNavigation to Navigation,
and AppNavigation/Authentication to AuthNavigation
- added a state fixture with the auth: INITIAL_STATE
- added a `createFixture` test util to make it easier to create test data

## Steps to Test
- ensure the app builds.


sciphergfx (changes requested):
Changes look good, however I'm unable to run the app. I think there's an issue on User.middleware file. Please confirm, thanks

sciphergfx (approved):
All good

PR #87, opened 2021-07-04
YOMA-480 - refactor(App): Error running app: Module AppRegistry is not a registered callable module

sciphergfx:
## Work Done
- imported App entry file directly into index.js
- Back navigation crashing device on auth screens (this icon needs to be removed from UI)


index.js
```diff
@@ -2,7 +2,7 @@ import React from 'react'
 import { AppRegistry } from 'react-native'
 import 'react-native-gesture-handler'
 import { config } from './config'
-import App from './src/modules/App'
+import App from './src/modules/App/App'
```
cameronolivier: this should be reverted.

src/modules/App/index.ts
```diff
@@ -1,6 +1,4 @@
-import App from './App'
-import * as appMiddleware from './App.middleware'
+import * as middleware from './App.middleware'
```
cameronolivier, line 2: have you removed the App export from the index?

cameronolivier (changes requested):

src/modules/App/index.ts
cameronolivier: I see this is what fixed the issue - however, I'm not happy with this workaround.  I don't think we're fixing anything here- we're simply avoiding it - which means it comes back to bite us at some point. 

I reverted your changes on the `index.js` <> `App.ts` - and it seems to be a bit better (although I still intermittently got that ui wait issue)

```js
import { RNLocalize, setI18nConfig } from 'locales/i18n'
import React, { useEffect } from 'react'
import FlashMessage from 'react-native-flash-message'

import AppNavigation from '../AppNavigation'

setI18nConfig()

const App = () => {
  useEffect(() => {
    RNLocalize.addEventListener('change', setI18nConfig)
    return () => {
      RNLocalize.removeEventListener('change', setI18nConfig)
    }
  }, [])

  return (
    <>
      <AppNavigation />
      <FlashMessage position="top" />
    </>
  )
}

export default App

```

```js
import React from 'react'
import { AppRegistry } from 'react-native'
import 'react-native-gesture-handler'
import { config } from './config'
import App from './src/modules/App'
import { persistor, store } from './src/redux/store'
import { Provider } from 'react-redux'
import { PersistGate } from 'redux-persist/integration/react'

const Root = () => (
  <Provider store={store}>
    <PersistGate loading={null} persistor={persistor}>
      <App />
    </PersistGate>
  </Provider>
)

AppRegistry.registerComponent(config.application.name, () => Root)

```
but I kept the App export in `App/index.ts`

I think this might a step towards a fix.

sciphergfx: > have you removed the App export from the index?

Yes I did, We're getting errors when exported as default in App/index.ts

sciphergfx: Picked up the lag as well, but I can update changes as above

cameronolivier (approved):

PR #88, opened 2021-07-05
YOMA-450 - refactor(User): about should correctly save any edits to backend

sciphergfx:
## Work Done
- Implemented save functionality for user biography
- Added tests where necessary

## Steps to Test
- Sign in into the app, and add a biography


src/modules/About/About.container.tsx
```diff
@@ -1,9 +1,16 @@
+import selector from 'modules/About/About.selector'
```
cameronolivier: fix this import please.

src/modules/About/About.container.tsx
```diff
@@ -1,9 +1,16 @@
...
 import { connect } from 'react-redux'
+import { Dispatch } from 'redux'
 
+import * as UserActions from '../User/User.reducer'
```
cameronolivier: import from the module index rather

src/modules/About/About.container.tsx
```diff
@@ -1,9 +1,16 @@
...
 import About from './About'
-import selector from './About.selector'
 
 const mapStateToProps = selector
-const mapDispatchToProps = null
-
+const mapDispatchToProps = (dispatch: Dispatch) => {
+  return {
+    onUpdateUserCredentials: (biography: string) => {
```
cameronolivier: this should be `onUpdateUser` as mentioned in PR for user data

However, I'd actually say it should be `onUpdateBiography` or `onUpdateAbout` as that's more specifically what we're updating here.

src/modules/User/User.middleware.test.ts
```diff
@@ -39,4 +52,115 @@ describe('modules/User/User.middleware', () => {
       expect(store.dispatch).toHaveBeenCalledWith(setUserCredentials(userData))
     })
   })
+  describe('updateUserCredentialsFlow', () => {
+    it('should correctly handle being called', async () => {
```
cameronolivier: this wouldn't be async

src/modules/User/User.middleware.test.ts
```diff
@@ -39,4 +52,115 @@ describe('modules/User/User.middleware', () => {
...
+        },
+      }
+      const create = createMiddlewareMock(jest, mockState)
+      const updateCredentials = {
+        biography: 'BIOGRAPHY',
+      }
+      const mockAction = {
+        payload: updateCredentials,
+      }
+      const action = updateUserCredentials(updateCredentials)
```
cameronolivier: this should be `updateUser` - we aren't dealing with credentials here.

src/modules/User/User.middleware.test.ts
cameronolivier: we need more test cases. 

1. should handle being called
2. should submit the updated user biography (testing that it calls our api middleware with the correct information)

src/modules/User/User.middleware.test.ts
```diff
@@ -39,4 +52,115 @@ describe('modules/User/User.middleware', () => {
...
+            onFailure: updateUserCredentialsFailure,
+            endpoint: userId,
+          }),
+          userUpdateCredentials,
+        ),
+      )
+    })
+  })
+  describe('updateUserCredentialsSuccessFlow', () => {
+    it('should correctly handle being called', () => {
```
cameronolivier, line 127: as above, we want a second test handling the expected success state.
I know this is a bit tedious, but we want to
1. test it's handling being called
2. test that the functionality of our middleware is succeeding as expected 
3. (optionally) test that it fails and we catch it in an expected manner
4. (any other specific use cases that might arise)

But we want to explicitly be testing specific things each time.

src/modules/User/User.middleware.ts
```diff
@@ -15,3 +31,57 @@ export const setUserOnAuthFlow: Middleware =
     }
     return result
   }
+
+export const updateUserCredentialsFlow: Middleware =
```
cameronolivier: As mentioned above -remove credentials.

src/modules/User/User.middleware.ts
```diff
@@ -15,3 +31,57 @@ export const setUserOnAuthFlow: Middleware =
...
+  ({ getState, dispatch }) =>
+  next =>
+  async action => {
```
cameronolivier: this isn't actually an async action as the api middleware handles the async aspect.

src/modules/User/User.middleware.ts
```diff
@@ -15,3 +31,57 @@ export const setUserOnAuthFlow: Middleware =
...
+    const result = next(action)
+    if (updateUserCredentials.match(action)) {
+      const state = getState()
+      const userId = selectUserIdFromUserCredentials(state)
```
cameronolivier: should just be named `selectUserId` and you can use a standard selector for that.

src/modules/User/User.middleware.ts
```diff
@@ -15,3 +31,57 @@ export const setUserOnAuthFlow: Middleware =
...
+            endpoint: userId,
+          }),
+          credentials,
+        ),
+      )
+    }
+    return result
+  }
+
+export const updateUserCredentialsSuccessFlow =
```
cameronolivier: again - remove credentials

src/modules/User/User.middleware.ts
```diff
@@ -15,3 +31,57 @@ export const setUserOnAuthFlow: Middleware =
...
+      const credentials = selectUserCredentialsFromUpdatePayload(action)
+      dispatch(setUserCredentials(credentials))
+      NavigationActions.navigate(HomeNavigationRoutes.Home)
+      // TODO: this should be handled by the notification module
+      notification('success', 'Details Updated')
+    }
+    return result
+  }
+
+export const updateUserCredentialsFailureFlow =
```
cameronolivier: again remove credentials

src/modules/User/User.middleware.test.ts
```diff
@@ -39,4 +52,115 @@ describe('modules/User/User.middleware', () => {
...
+      }
+      const action = updateUserCredentials(updateCredentials)
+      // @ts-ignore
+      const { invoke, store } = create(SUT.updateUserCredentialsFlow)
+
+      // when ... we respond to the updateUserCredentials action
+      await invoke(action)
+
+      // then ...validate updateUserCredentialsFlow
+      const userId = selectUserIdFromUserCredentials(mockState)
```
cameronolivier: you don't need to do this - you have the USER_ID value already above.

src/modules/User/User.middleware.test.ts
```diff
@@ -39,4 +52,115 @@ describe('modules/User/User.middleware', () => {
...
+          organisation: 'ORGANISATION',
+          createdAt: 'CREATED_AT',
+          lastLogin: 'LAST_LOGIN',
+        },
+      }
+      const create = createMiddlewareMock(jest, mockState)
+      const updateCredentials = {
+        biography: 'BIOGRAPHY',
+      }
+      const mockAction = {
```
cameronolivier, line 100: this test is way too complicated. 

we don't need the mockAction. 
You can simply set the action payload here:
```js
const action = updateUserCredentials('A PAYLOAD')
```
or if you want to be more specific, you could do: 
```js
const action = updateUserCredentials({biography: 'A BIOGRAPHY'})
```

src/modules/User/User.middleware.test.ts
```diff
@@ -39,4 +52,115 @@ describe('modules/User/User.middleware', () => {
...
+      const action = updateUserCredentials(updateCredentials)
+      // @ts-ignore
+      const { invoke, store } = create(SUT.updateUserCredentialsFlow)
+
+      // when ... we respond to the updateUserCredentials action
+      await invoke(action)
+
+      // then ...validate updateUserCredentialsFlow
+      const userId = selectUserIdFromUserCredentials(mockState)
+      const userUpdateCredentials = selectUserUpdateCredentials(mockAction)(mockState)
```
cameronolivier: we shouldn't need this.

src/modules/User/User.middleware.test.ts
```diff
@@ -39,4 +52,115 @@ describe('modules/User/User.middleware', () => {
...
+      expect(store.dispatch).toHaveBeenCalledWith(
+        ApiActions.apiRequest(
+          mergeRight(ApiUserConstants.USERS_EDIT_CONFIG, {
+            onSuccess: updateUserCredentialsSuccess,
+            onFailure: updateUserCredentialsFailure,
+            endpoint: userId,
```
cameronolivier: we can just set this, we don't need to use the util - we already have it above in the mockState: 
```js
id: 'USER_ID',
```

src/modules/User/User.middleware.test.ts
cameronolivier: So please follow the same testing flow (as detailed above) as you've done in all the other middleware.

src/modules/User/User.middleware.test.ts
```diff
@@ -39,4 +52,115 @@ describe('modules/User/User.middleware', () => {
...
+          }),
+          userUpdateCredentials,
```
cameronolivier: we can just set this again in text - the more explicit we are with tests the easier it is to see what's going on, and understand what we're trying to test. 

So if you take my example above, this will simple either be a string:
```js
 'A PAYLOAD'
```
or an object:
```js
{biography: 'A BIOGRAPHY'}
```

src/modules/User/User.selector.ts
```diff
@@ -1,7 +1,21 @@
 import { createSelector } from '@reduxjs/toolkit'
-import { propOr } from 'ramda'
+import { applySpec, mergeDeepRight, pipe, prop, propOr, __ } from 'ramda'
 import { RootState } from 'redux/redux.types'
 
 export const selectUserState = (state: RootState) => state.user
 
 export const selectBiography = createSelector(selectUserState, propOr('', 'biography'))
+export const selectUserUpdateCredentials = (action: any) =>
```
cameronolivier: the selector isn't the place for a util to extract data from the api response. 

only functions (selectors) that act on state data should be here.

src/modules/User/User.selector.ts
```diff
@@ -1,7 +1,21 @@
...
+  createSelector(
+    selectUserState,
+    pipe(
```
cameronolivier: as an aside, I'm unsure why you're specifically only returning these values from the api call? I'd suggest you simply update the user object with the user data you get back - which is simple enough to extract - and removes the need for this map/apply spec.

src/modules/User/User.middleware.ts
```diff
@@ -15,3 +31,57 @@ export const setUserOnAuthFlow: Middleware =
...
+
+export const updateUserCredentialsSuccessFlow =
+  ({ notification }: { notification: typeof showSimpleMessage }): Middleware =>
+  ({ dispatch }) =>
+  next =>
+  action => {
+    const result = next(action)
+
+    if (updateUserCredentialsSuccess.match(action)) {
+      const credentials = selectUserCredentialsFromUpdatePayload(action)
```
cameronolivier: I think all you need to do here is get the data from the payload and dispatch off a `setUser` action with the user object.

src/modules/User/User.types.ts
```diff
@@ -0,0 +1,15 @@
+import { ApiMetaResponse, UserResponse } from 'modules/Auth/Auth.types'
+
+export interface UpdateUserCredentialsResponse {
+  data: UserResponse
+  meta: ApiMetaResponse
+}
+export type UpdateUserCredentialsFailureResponse = string
+
+export interface UserCredentialsPayload {
```
cameronolivier: this is not the credentials, this is the user.

cameronolivier (changes requested):
quite a few changes required.

src/modules/User/User.selector.ts
sciphergfx: If I understand you correctly, updating reducer state during this operation will persist the data

src/modules/User/User.middleware.test.ts
sciphergfx: noted

src/modules/User/User.middleware.ts
```diff
@@ -15,3 +31,57 @@ export const setUserOnAuthFlow: Middleware =
...
+
+export const updateUserCredentialsFlow: Middleware =
+  ({ getState, dispatch }) =>
+  next =>
+  async action => {
+    const result = next(action)
+    if (updateUserCredentials.match(action)) {
+      const state = getState()
+      const userId = selectUserIdFromUserCredentials(state)
+      const credentials = selectUserUpdateCredentials(action)(state)
```
cameronolivier:
```js
const user = selectUser(state)
const userProfile = pick(['lastName', ...])(user)
const payload = mergeRight(userProfile, action.payload)
```
or as a util:
```js
const createUserProfilePayload = state => payload => pipe(
 selectUser, 
 pick(['lastName', ...]),
 mergeLeft(action.payload)
)(state)
```

src/modules/About/About.container.tsx
sciphergfx: this change cuts across 3 branches, I think it makes sense to implement on base x branch (YOMA-436) before merge with develop.

src/modules/User/User.middleware.test.ts
```diff
@@ -39,4 +42,116 @@ describe('modules/User/User.middleware', () => {
...
+      }
+
+      const mockNotification = jest.fn()
+      const action = updateUserSuccess(mockResponseData)
+      // @ts-ignore
+      const { store, invoke, next } = create(SUT.updateUserSuccessFlow({ notification: mockNotification }))
+      // when ... we respond to the updateUserSuccess action
+      invoke(action)
+      // then ...validate updateUserSuccessFlow
+      const credentials = selectUserFromUpdatePayload(action)
```
cameronolivier: also credentials rename to user

src/modules/User/User.middleware.ts
```diff
@@ -15,3 +21,57 @@ export const setUserOnAuthFlow: Middleware =
...
+
+export const updateUserSuccessFlow =
+  ({ notification }: { notification: typeof showSimpleMessage }): Middleware =>
+  ({ dispatch }) =>
+  next =>
+  action => {
+    const result = next(action)
+
+    if (updateUserSuccess.match(action)) {
+      const credentials = selectUserFromUpdatePayload(action)
```
cameronolivier: please rename to 'user'

cameronolivier: This fn should be called `extractUser`, as this one is specifically doing that. On second thought, I think `extractUserFromUserUpdateSuccess` is the best (I can come up with) for now - as we would have a similar (but different) `extract` function for getting the user object from the Auth success.

src/modules/User/User.middleware.ts
```diff
@@ -15,3 +21,57 @@ export const setUserOnAuthFlow: Middleware =
...
+
+export const updateUserFlow: Middleware =
+  ({ getState, dispatch }) =>
+  next =>
+  action => {
+    const result = next(action)
+    if (updateUser.match(action)) {
+      const state = getState()
+      const userId = selectUserId(state)
+      const user = extractUser(action)(state)
```
cameronolivier: sorry to be a stickler, but how we name things is important. 

this function isn't really extracting the user so much as it's preparing the data for the api patch. right? 

So you'd want to call it something that reflects that. Something like `prepareUserPatch` or `prepareUserPatchData` maybe?

cameronolivier (changes requested):

src/modules/User/User.middleware.test.ts
sciphergfx: So I'm using mockAction in two different places, I think it make sense it remains a variable

src/modules/User/User.middleware.ts
sciphergfx: I understand, but as agreed previously, went with 'extractUser' to name this util fn to make our lives easy. I think we spoke about this in a meeting, perhaps a different function was referenced? but happy to change it.

src/modules/User/User.middleware.test.ts
sciphergfx: there are 3 expects in this test that captures the functions in this middleware.
```js
      expect(next).toHaveBeenCalledWith(action)
      expect(store.dispatch).toHaveBeenCalledWith(setUser(user))
      expect(mockNotification).toHaveBeenCalled()
```
We've had just one 'it' test case for most *SuccessFlow app-wide. I think we need to standardise this across

sciphergfx: I'll break / add more cases on this @cameronolivier

src/modules/User/User.utils.ts
```diff
@@ -0,0 +1,5 @@
+import { path, pipe } from 'ramda'
+
+export const selectUserFromLoginPayload = pipe(path(['payload', 'data', 'data', 'user']))
```
cameronolivier: None of these need a `pipe`. Pipe is only required if we have more than 1 function - as in we're piping our functions down the line. As these are all single functions, calling the function directly is perfectly ok. 

As these are new functions, they should be named with `extract` not `select` as discussed.

src/modules/User/User.utils.ts
```diff
@@ -0,0 +1,5 @@
...
+export const extractUserFromUserUpdateSuccess = pipe(path(['payload', 'data', 'data']))
+export const selectUserId = pipe(path(['user', 'id']))
```
cameronolivier: This one - should it not be a selector? It looks like it's taking state? Or what data do we pass into it? If we're using it to select the user ID from state, then we should create a selector for it (it is a much-needed selector)

src/modules/User/User.utils.ts
```diff
@@ -0,0 +1,5 @@
+import { path, pipe } from 'ramda'
+
+export const selectUserFromLoginPayload = pipe(path(['payload', 'data', 'data', 'user']))
+export const extractUserFromUserUpdateSuccess = pipe(path(['payload', 'data', 'data']))
```
cameronolivier: This needs to be tested.

src/modules/User/User.utils.ts
cameronolivier: I see we are using it to get the selector out of state - so it is a selector-  and you can then simply use it in the middleware as you currently are.

It'll need to be tested.

src/modules/User/User.selector.ts
```diff
@@ -0,0 +1,6 @@
+import { createSelector } from '@reduxjs/toolkit'
+import { propOr } from 'ramda'
+import { RootState } from 'redux/redux.types'
+
+export const selectUser = (state: RootState) => state.user
+export const selectBiography = createSelector(selectUser, propOr('', 'biography'))
```
cameronolivier: as mentioned below- the `selectId` util should go here.

src/modules/User/User.selector.test.ts
```diff
@@ -0,0 +1,53 @@
+import { USER_RESPONSE } from './../Profile/Profile.constants'
+import { INITIAL_STATE } from './User.reducer'
+import * as SUT from './User.selector'
+
+describe('modules/User/User.selector', () => {
+  describe('selectBiography ', () => {
```
cameronolivier: I think these are changes set to come in in YOMA-442 as well (but both of these point at develop). 

So, I'll repeat my comment - should use `rootStateFixture()` here, as your state object is being flagged by typescript as it isn't an accurate representation of the expected redux state (it lacks the auth endpoint).

src/modules/About/About.selector.test.ts
```diff
@@ -0,0 +1,52 @@
...
+      // given ...
+      const state = {
+        user: INITIAL_STATE,
+      }
+      // when ... we call the selector
+      const result = SUT.default(state)
+      // then ... should return result as expected
+      expect(result).toEqual({ biography: '' })
+    })
+    it('should handle empty state if props not available', () => {
```
cameronolivier: As mentioned in the other PR comments, we can remove this, and should be properly mocking state comprehensively. See comments on YOMA-442 PR first.

cameronolivier (changes requested):

src/components/SocialRegistration/SocialRegistration.tsx
```diff
@@ -1,61 +1,15 @@
-import { GoogleSignin, statusCodes } from '@react-native-google-signin/google-signin'
 import React from 'react'
-import { LoginManager } from 'react-native-fbsdk'
 
 import SocialButton from '../SocialButton'
 import { SocialVariants } from '../SocialButton/SocialButton.types'
 import Spacer from '../Spacer'
 
 const SocialRegistration = () => {
-  const facebookSignIn = () => {
```
sciphergfx: Took this out because of test related errors. SSO PR should fill this changes. I also think we should revert back to merge instead of 'squashing' before merge, makes refactoring a bottleneck.

src/modules/About/About.container.tsx
```diff
@@ -1,9 +1,16 @@
...
+import { actions as UserActions } from '../User'
 import About from './About'
 import selector from './About.selector'
 
 const mapStateToProps = selector
-const mapDispatchToProps = null
-
+const mapDispatchToProps = (dispatch: Dispatch) => {
+  return {
+    onPatchUserData: (biography: string) => {
```
cameronolivier: OPTIONAL: From a naming pov, I'd say this should reflect where it's being called - so it's an `onSaveAbout` or onUpdateBiography` - something like that. At this level of abstraction, the code is only concerned with doing something when the user clicks to save their about data. Down the line we might change what action we're calling and not touch the user endpoint - or be saving to a biography credential (idk) - my point is that this should reflect what the call to action is addressing in its naming convention.

The action could easily be patchUser (but updateUser is fine too).

src/modules/User/User.middleware.ts
```diff
@@ -1,17 +1,78 @@
...
+import { extractUserFromLoginPayload, extractUserFromUserUpdateSuccess } from './User.utils'
 
 export const setUserOnAuthFlow: Middleware =
   ({ dispatch }) =>
   next =>
   action => {
     const result = next(action)
     if (authLoginSuccess.match(action)) {
-      const user = selectUserFromLoginPayload(action)
+      const credentials = extractUserFromLoginPayload(action)
```
cameronolivier: credentials is the incorrect term - it should be "user"

src/modules/User/User.selector.ts
```diff
@@ -1,6 +1,7 @@
 import { createSelector } from '@reduxjs/toolkit'
-import { propOr } from 'ramda'
+import { prop } from 'ramda'
 import { RootState } from 'redux/redux.types'
 
 export const selectUser = (state: RootState) => state.user
-export const selectBiography = createSelector(selectUser, propOr('', 'biography'))
+export const selectBiography = createSelector(selectUser, prop('biography'))
+export const selectUserId = createSelector(selectUser, prop('id'))
```
cameronolivier: OPTIONAL, as it's a small win, but this could simply be `selectId` - as it's part of the user object, would make sense. As is still works though.

src/modules/About/About.tsx
```diff
@@ -11,10 +12,11 @@ import styles from './About.styles'
 
 interface Props {
   biography: string
+  onPatchUserData: (biography: UserPayload) => void
   navigation: StackNavigationProp<HomeNavigatorParamsList, HomeNavigationRoutes.About>
 }
 
-const About = ({ navigation, biography }: Props) => {
+const About = ({ navigation, onPatchUserData, biography }: Props) => {
```
cameronolivier: I'm really not happy with the naming conventions used - but won't let that block this going in.

In our About component, we're calling the 'about' text: 
summary (in `summary` and `setSummary`), biography, about, and (to a degree 'userData'). In a single component. So that's not great.

Also the callback functions within a component should be named according to what's happened within the component - not what needs to happen down the line.

At the moment we have the almost awkward convention that in the About component, we've named the callback to update the user's biography `onPatchUserData`, but the actual action we're using to 'patch' the user data is: `updateUser`. 

I think that this should be named something like `onBiographySave`. 

As said - I'm (not super happy) - but ok to let this in as-is considering the code works and it's tested. But we need to be better on these things going forward.

cameronolivier (dismissed):
All optional changes, and I think 1 question, but happy to approve as it works.

src/modules/User/User.middleware.test.ts
```diff
@@ -1,74 +1,266 @@
...
 describe('modules/User/User.middleware', () => {
-  describe('setUserOnAuthFlow', () => {
-    it('should correctly handle being called', () => {
-      // given ... the authLogin action is fired
-      const create = createMiddlewareStub(jest)
-      const user = defaultUserLoginResponseData
-      const action = authLoginSuccess(user)
-      // @ts-ignore
-      const { store, invoke, next } = create(SUT.setUserOnAuthFlow)
+  it('should correctly handle being called', () => {
```
cameronolivier: we need a describe for `setUserOnAuthFlow` above here.

src/modules/User/User.middleware.ts
```diff
@@ -1,25 +1,58 @@
 import { authLoginSuccess } from 'modules/Auth/Auth.reducer'
+import { HomeNavigationRoutes } from 'modules/HomeNavigation/HomeNavigation.types'
 import { mergeRight } from 'ramda'
 import { Middleware } from 'redux'
+import { showSimpleMessage } from 'utils/error'
 
 import { actions as ApiActions, utils as ApiUtils } from '../../api'
-import { constants as ApiUsersConstants } from '../../api/users'
-import { fetchUserCredentials, fetchUserCredentialsFailure, fetchUserCredentialsSuccess, setUser } from './User.reducer'
+import { constants as ApiUserConstants, constants as ApiUsersConstants } from '../../api/users'
```
cameronolivier: this is being imported twice as slightly different constants. We need to fix it.

src/modules/User/User.middleware.ts
```diff
@@ -41,3 +74,33 @@ export const fetchUserCredentialsFlow: Middleware =
...
+  ({ notification }: { notification: typeof showSimpleMessage }): Middleware =>
+  ({ dispatch }) =>
+  next =>
+  action => {
+    const result = next(action)
+
+    if (updateUserSuccess.match(action)) {
+      const user = extractUserFromUserUpdateSuccess(action)
+      dispatch(setUser(user))
+      Navigation.navigate(HomeNavigationRoutes.Home)
```
cameronolivier, line 88: we should pass Navigation in as a dependency as well - otherwise it's going to be hard to mock it

src/modules/User/User.middleware.test.ts
```diff
@@ -1,74 +1,266 @@
...
+    const action = updateUserSuccess(mockResponseData)
+    // @ts-ignore
+    const { store, invoke } = create(SUT.updateUserSuccessFlow({ notification: mockNotification }))
+    // when ... we respond to the updateUserSuccess action
+    invoke(action)
+    // then ...validate setUser
+    const user = extractUserFromUserUpdateSuccess(action)
+    expect(store.dispatch).toHaveBeenCalledWith(setUser(user))
+  })
+  it('should correctly send notification to the user ', () => {
```
cameronolivier: we also need to test that we're navigating correctly.

cameronolivier: let's add a todo here - and we can add when we touch the code at a later point - as we're going to be updating the SSO stuff anyway

src/modules/User/User.middleware.ts
cameronolivier: but... let's wait if it's working now as we're updating the SSO anyway - maybe add a TODO in the code so we address it next time.

src/modules/User/User.types.ts
```diff
@@ -1,3 +1,19 @@
+import { ApiMetaResponse } from 'modules/Auth/Auth.types'
+
+export interface UpdateUserResponse {
+  data: UserResponse
+  meta: ApiMetaResponse
+}
+export type UpdateUserFailureResponse = string
+
+export interface UserPayload {
```
cameronolivier, line 9: DON'T DO NOW. (but we should refactor at some point and do this going forward)
let's not do this now - but a better way to do this - in a way that we ensure better type safety is to rather have everything required - and pick the properties that we're updating each time:
```js
export interface UserPayload {
  firstName: string
  lastName: string
  phoneNumber: string
  countryAlpha2: string
  biography: string
}
```
then when we use it:
```js
const onSaveBiography = (biography: Pick<UserPayload, 'biography'>) => ...
```
 [typescript Pick docs](https://www.typescriptlang.org/docs/handbook/utility-types.html#picktype-keys)

And only if we ever are updating and aren't sure what exactly we're updating, we could then do something like: 
```js
const onUpdateRandomUserInformation = (user: Partial<UserPayload>) => ... 
```

cameronolivier: It's not SUPER important right now as EVERYTHING is a string... so it doesn't really matter. but we could also have an empty object (as everything is optional) which isn't what we want.

cameronolivier (dismissed):
There are issues here. It needs to be fixed at some point. But in the interests of a level of progress, I'm approving as none seem to be bugs. 

We have a circular import that's bothering me though - I think it'll be fixed if we pull the Navigation stuff out of here and import in middleware, but let's keep an eye on it and otherwise tackle that with SSO refactor.

src/modules/User/User.middleware.ts
cameronolivier: I just realised this isn't part of SSO work. we should add a TODO for this if we don't do it now.

cameronolivier (dismissed):

cameronolivier (approved):

PR #91, opened 2021-07-08
YOMA-468  - feat(Auth): implement refresh token in auth flow

cameronolivier:
## Scope [YOMA-468](https://yomaworld.atlassian.net/browse/YOMA-468)
- authenticate a logged out user with the refresh token if possible.

## Work Done
- added middleware to get refresh token from storage and attempt to authorize the user
- added middleware to catch an api error and check if the user is unauthorized and authorize if they are.
- added middleware to log the user out if there is no refresh token present or if reauth fails. 
- passed the failure action to the Error module - it will filter out unauthorized errors and only send the failure action if not an unauthorized error. (I'm not sure I like this)

### Additional:
- fixing some circular dependency warnings


sciphergfx (dismissed):
I see some the auth names have been revised, otherwise all good here

sciphergfx (approved):
All good

PR #92, opened 2021-07-11
YOMA-437 - feat(profile) - save edited user profile data

sciphergfx:
## Work Done
- implemented actions to save user data
- callback to set default value on dropdown component (fixed an existing bug here)
- Added clear user action in reducer
- Added test for this
- Removed spinner in profile screen (We need to add a loader to our redux, I think a new ticket on this will be better as we need it in other parts of the app)

## Steps to Test
sign into application, navigate to profile and update user data.


src/modules/Profile/Profile.form.tsx
```diff
@@ -38,12 +37,12 @@ const ProfileForm = forwardRef(({ user }: Props, ref) => {
...
       enableReinitialize
       validationSchema={profileValidationSchema}
-      onSubmit={() => {}}
+      onSubmit={onPatchUserData}
     >
       {formikHandlers => {
         return (
           <>
-            <Spinner visible={formikHandlers.isSubmitting} />
+            {/* <Spinner visible={formikHandlers.isSubmitting} /> */}
```
sciphergfx, line 42: Visible prop value should be from redux, I think it's best tackled on a different ticket as we'll need loader in other parts

src/modules/Profile/Profile.container.ts
```diff
@@ -11,6 +12,9 @@ const mapDispatchToProps = (dispatch: Dispatch) => {
     onLogoutUser: () => {
       dispatch(AuthActions.authLogout())
     },
+    onPatchUserData: (user: any) => {
```
cameronolivier: as mentioned in the other PR, this should be named according to what we're doing here (so update or saving the user's information - I think we can generally remove 'data' - as 'saveUser' is understood to mean 'save user's data'.

I'm happy to call our actions 'patchUser' instead of 'updateUser', but I think update works fine for now.

src/modules/Profile/Profile.form.tsx
cameronolivier: Agreed -all loading states stopped/started by side effects/api calls should be handled within redux.

Any local loading states should be handled internally.

src/modules/Profile/Profile.tsx
```diff
@@ -15,22 +15,28 @@ import styles from './Profile.styles'
 
 interface Props {
   onLogoutUser: () => void
-  user: UserResponse
+  onPatchUserData: (user: any) => void
+  user: any
```
cameronolivier: why revert on the typing?

cameronolivier: if it's only a partial user response, then we could use `Partial<UserResponse>` - but I suspect that UserResponse isn't correct as that's the response from the API, so we should type it out specifically for the fields changing on the Profile form.

src/modules/Profile/Profile.tsx
```diff
@@ -15,22 +15,28 @@ import styles from './Profile.styles'
...
   navigation: StackNavigationProp<HomeNavigatorParamsList, HomeNavigationRoutes.Profile>
 }
 
-const Profile = ({ navigation, onLogoutUser, user }: Props) => {
-  const [userResponse] = useState<UserResponse>(user)
+const Profile = ({ navigation, onLogoutUser, onPatchUserData, user }: Props) => {
+  const [userData] = useState<UserResponse>(user)
```
cameronolivier: where possible, let's remove 'data' or 'info' as 9/10 it's superfluous.

cameronolivier: Also if we never update the `userData`, then it's essentially the same as the `user` prop, so I'm not sure why we would want to create a new variable.

src/modules/User/User.middleware.ts
```diff
@@ -31,6 +31,7 @@ export const updateUserFlow: Middleware =
     if (updateUser.match(action)) {
       const state = getState()
       const userId = selectUserId(state)
+      const patchPayload = extractUserPatchPayload(action.payload)
```
cameronolivier: I'd suggest this be userPayload. 'patch' here says nothing about the data. it's the same as `postPayload` or `getPayload` which means nothing really. 

I think if we remove the 'patch' part altogether across our naming (unless we adopt it in our actions in place of 'update'), it'll really help our naming. 

```js
const user = extractUserfromUpdateUserPayload(action.payload)
```
should be perfectly ok.

cameronolivier (changes requested):
Mostly naming convention issues. Will test the code once this is done.

src/modules/Profile/Profile.tsx
```diff
@@ -2,31 +2,39 @@ import { StackNavigationProp } from '@react-navigation/stack'
...
 
-const Profile = ({ navigation, onLogoutUser, user }: Props) => {
+const Profile = ({ navigation, onLogoutUser, onProfileSave, user }: Props) => {
   const [userResponse] = useState<UserResponse>(user)
+  const [formState, setFormState] = useState<ProfileFormState>(null)
+
   const { t } = useTranslation()
-  const childRef = useRef<FormikProps<FormikValues>>()
+
+  const handleProfileForm = () => {
```
cameronolivier, line 29: we might need to wrap this in a useCallback - but let's see - as long as we're not getting into any weird ui things we should be ok.

cameronolivier (commented):
Going to wait for 450 to be in develop and then I'll give it a run through - happy with things so far.

cameronolivier (commented):
Happy with the code, ~but I tried to upload an image and it failed to open the camera or anything~. I realise there is another task for the image.

cameronolivier (dismissed):

cameronolivier (dismissed):

cameronolivier (approved):

PR #95, opened 2021-07-12
YOMA-488 - refactor(LargeHeader): fixing navigate back

cameronolivier:
## Scope [YOMA-488](https://yomaworld.atlassian.net/browse/YOMA-488)

- fix the back buttons on register and login screen.

## Work Done
- removed the current handlers and set the back button to `navigation.back` in `LargeHeader`
- removed unused `HeaderContainer` component
- removed the LargeHeader component from `component/index.ts`

## Steps to Test
- Open the app on the Landing screen, navigate to Login > and press back, then do the same with Register.


src/components/LargeHeader/LargeHeader.tsx
```diff
@@ -25,17 +20,9 @@ const LargeHeader = ({ circleImage, circleImageStyle, backgroundColor, navigatio
     setStyle(StyleSheet.flatten([backgroundColorStyle, styles.container]))
   }, [backgroundColor])
 
-  useEffect(() => {
-    const backHandler = BackHandler.addEventListener('hardwareBackPress', () => {
```
sciphergfx, line 29: Why are we removing native navigation support?

sciphergfx: We can still run the back function when native back event is fired
```js
 useEffect(() => {
    const backHandler = BackHandler.addEventListener('hardwareBackPress', () => {
      navigation.goBack()
    })
    return () => backHandler.remove()
  }, [navigation])
```

sciphergfx: Just tested, I see react-navigation now supports that out of the box, all good!

sciphergfx (approved):
All good here

PR #96, opened 2021-07-15
YOMA-438 - refactor(Profile): save new profile image

sciphergfx:
## Work Done
- Support for profile image upload
- Show error pop up when image upload fails


src/modules/Profile/Profile.tsx
```diff
@@ -31,6 +33,11 @@ const Profile = ({ navigation, onLogoutUser, onProfileSave, user }: Props) => {
       onProfileSave(formState.values)
     }
   }
+  const handleProfilePhoto = () => {
```
cameronolivier: please move this to middleware. 

fire an action to save the image. then handle what you're doing in the util in the middleware - then catch the success and fire the onPhotoSave action.

cameronolivier (changes requested):

src/redux/store.ts
```diff
@@ -18,7 +19,14 @@ const store = configureStore({
...
     getDefaultMiddleware({
       serializableCheck: {
-        ignoredActions: [FLUSH, PAUSE, PERSIST, REGISTER, apiActions.apiRequest.type],
+        ignoredActions: [
+          FLUSH,
+          PAUSE,
+          PERSIST,
+          REGISTER,
+          apiActions.apiRequest.type,
+          userActions.uploadUserPhotoSuccess.type,
```
sciphergfx, line 28: form comes with unserialisable values

src/modules/User/User.middleware.test.ts
```diff
@@ -218,52 +223,184 @@ describe('modules/User/User.middleware', () => {
...
+      const FormDataMock = function (this: any) {
+        this.formData = 'FORM_DATA'
+        this.append = jest.fn()
+        return this.formData
+      } as any
+      const create = createMiddlewareStub(jest)
+      // given ...
+
+      //TODO: Fix mock for image capture
+      const capturedProfileImageMock = () =>
```
sciphergfx, line 287: Struggling to mock this

PR #99, opened 2021-07-20
YOMA-452 - feat(Experience): Experience - save new experience to backend

sciphergfx:
## Work Done 
- Moved fetchCredentials to userCredentials module
- Added state clear actions to resetAppData
- Added util to extract error messages
- Added callback to set default value on dropdown component (fixed an existing bug here)
- Added searchable to all dropdown with long list for better UX (Experience Module)
- Moved fetchCredentials to it’s own credentials middleware as it has a different data structure to users,
- Added tests where necessary

Blocks - 
- “isWorkingHere prop” not included in jobs endpoint 
- requestVerification is useful when a new organisation is created by the user, current mobile UI doesn't make any provision for this.
- fetch credentials doesn't return skills for job


src/modules/Credentials/Credentials.middleware.test.tsx
```diff
@@ -0,0 +1,39 @@
+import { mergeRight } from 'ramda'
```
cameronolivier: there is a Credentials endpoint on swagger - that isn't the same as the UserCredentials - so I don' think it's a good idea to have a module here called Credentials that is the UserCredentials.

I'd suggest moving this back into the User module. 
Alternatively, rename this to UserCredentials module - and then maintain that naming convention throughout.

src/modules/App/App.middleware.ts
```diff
@@ -12,6 +13,7 @@ export const appResetFlow: Middleware =
 
     if (resetAppData.match(action)) {
       dispatch(AuthActions.clearAuth())
+      dispatch(UserActions.clearUser())
```
cameronolivier: we should add an action to clearUserCredentials (presuming you're not going to move it back to User module), Organisations and Skills here as well.

src/modules/Credentials/Credentials.middleware.tsx
```diff
@@ -0,0 +1,164 @@
...
+import {
+  createCredentials,
+  setCredentials,
+  updateCredentials,
+  updateCredentialsFailure,
+  updateCredentialsSuccess,
+} from './Credentials.reducer'
+import { extractCredentialsFromHydratePayload } from './Credentials.utils'
+
+export const setCredentialsFlow: Middleware =
```
cameronolivier: we need to stick to a naming convention - half of these are UserCredentials, the other have is just Credentials. 

considering the above (about there being another credentials endpoint already, I'd adjust all of these to `UserCredentials`.

cameronolivier: (actions included)

src/modules/Credentials/Credentials.middleware.tsx
```diff
@@ -0,0 +1,164 @@
...
+  ({ dispatch }) =>
+  next =>
+  action => {
+    const result = next(action)
+    if (authLoginSuccess.match(action)) {
+      const user = extractCredentialsFromHydratePayload(action)
```
cameronolivier: this should be `userCredentials`. If you're going to copy-paste code that's 100% but please make sure to update the naming of your variables to make sense in their new context.

cameronolivier: and would be `extractUserCredentials...`

src/modules/Credentials/Credentials.middleware.tsx
```diff
@@ -0,0 +1,164 @@
...
+            onFailure: fetchUserCredentialsFailure,
+          }),
+          action.payload,
+        ),
+      )
+    }
+    return result
+  }
+
+export const createCredentialsFlow: Middleware =
```
cameronolivier: this middleware and the next one (updateCredentialsFlow) looks like it's exactly the same - so we probably only need 1 of these.

src/modules/Credentials/Credentials.middleware.tsx
```diff
@@ -0,0 +1,164 @@
...
+  }: {
+    notification: typeof showSimpleMessage
+    navigation: typeof Navigation
+  }): Middleware =>
+  _store =>
+  next =>
+  action => {
+    const result = next(action)
+
+    if (updateCredentialsSuccess.match(action)) {
```
cameronolivier: do we not need to handle any of the data coming back with updateCredentialsSuccess - like save it to the store?

src/modules/Credentials/Credentials.middleware.tsx
```diff
@@ -0,0 +1,164 @@
...
+    const result = next(action)
+
+    if (updateCredentialsFailure.match(action)) {
+      // TODO: this should be handled by the notification module
+      notification('danger', 'An error occurred.', 'Oops something went wrong! Please try again.')
+    }
+    return result
+  }
+
+export const createCredentialsSuccessFlow =
```
cameronolivier: this is the same as updateCredentialsSuccessFlow, so there's some duplication here.

src/modules/Credentials/Credentials.middleware.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+    if (updateCredentialsSuccess.match(action)) {
+      navigation.navigate(HomeNavigationRoutes.Home)
+      // TODO: this should be handled by the notification module
+      notification('success', 'Details Saved')
+    }
+    return result
+  }
+
+export const createCredentialsFailureFlow =
```
cameronolivier: likewise - the same as update... failure

src/modules/Credentials/Credentials.reducer.tsx
```diff
@@ -0,0 +1,26 @@
...
+export const fetchUserCredentials = createAction(`${name} fetchUserCredentials`)
+export const fetchUserCredentialsSuccess = createAction<any>(`${name} fetchUserCredentialsSuccess`)
+export const fetchUserCredentialsFailure = createAction<string>(`${name} fetchUserCredentialsFailure`)
+
+export const updateCredentials = createAction<Credential>(`${name} updateCredentials`)
+export const updateCredentialsSuccess = createAction(`${name} updateCredentialsSuccess`)
+export const updateCredentialsFailure = createAction<string>(`${name} updateCredentialsFailure`)
+
+const CredentialsReducer = createReducer(INITIAL_STATE, builder => {
+  builder.addCase(setCredentials, (state, action) => state.concat(action.payload))
```
cameronolivier: each credential should be storing their own data. 

so Experience should have its own reducer storing the user's experience, which we should be pulling out of the response we get for credentials- so I don't see Credentials itself storing the data. 

This is why I set it up the way I did - we're simply fetching the data - and it's up to each of the modules to extract the data they need from the payload.

Then, I've mentioned it before - but we need to be normalising our data as such: 
https://redux.js.org/usage/structuring-reducers/normalizing-state-shape
And redux toolkit has some great examples: https://redux-toolkit.js.org/usage/usage-guide#managing-normalized-data
I'd suggest we use normalizr library: https://github.com/paularmstrong/normalizr
So for the endpoints where we have lists of data, we want something like this: 
```ts
experience: {
  entities: {
    experienceId1: {....},
    experienceId2:{....},
    ...
  }
 ids: ['experienceId1', 'experienceId2', ...],
}
```

src/modules/Credentials/Credentials.types.ts
```diff
@@ -0,0 +1,49 @@
+export type QualificationResponsePayload = {
```
cameronolivier: I'd suggest the Qualification items move to the Experience module - presuming that's where they should exist.

src/modules/Credentials/Credentials.utils.ts
```diff
@@ -0,0 +1,3 @@
+import { path } from 'ramda'
+
+export const extractCredentialsFromHydratePayload = path([]) //TODO: add path
```
cameronolivier: I'd think this shoudl rather be named `..FromFetchUserCredentialsPayload`? - because this isn't coming in from the Hydrate payload - but on the Success action for FetchUserCredentials: This one `CredentialsActions.fetchUserCredentials()` -> which should fire `fetchUserCredentialsSuccess` with the payload.

src/modules/Experience/Experience.container.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { connect } from 'react-redux'
+import { RootState } from 'redux/redux.types'
+
+import { actions as OrganisationActions } from '../Organisations'
+import { actions as SkillsActions } from '../Skills'
+import Experience from './Experience'
+
+const mapStateToProps = (state: RootState) => {
+  return {
+    qualifications: [],
```
cameronolivier: I'm presuming 'qualifications' == experience items?

src/modules/Experience/Experience.selector.tsx
```diff
@@ -0,0 +1,6 @@
+import { createSelector } from '@reduxjs/toolkit'
```
cameronolivier: as mentioned elsewhere, I think we should have middleware that gets the experience/qualifications out of the credentials payload, normalizes it and stores it in the Experience reducer. 

Then we have an Experience.selector that selects that data from state.

src/modules/Experience/Experience.container.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { selector as OrganisationsSelectors } from 'modules/Organisations'
+import { selector as SkillsSelectors } from 'modules/Skills'
+import { connect } from 'react-redux'
+import { RootState } from 'redux/redux.types'
+
+import { actions as OrganisationActions } from '../Organisations'
+import { actions as SkillsActions } from '../Skills'
+import Experience from './Experience'
+
+const mapStateToProps = (state: RootState) => {
```
cameronolivier: you should create a selector file that handles this mapping as we've done elsewhere, then it should just look like: 
```ts
const mapStateToProps = selector
```

cameronolivier: this selector should be tested.

src/modules/Experience/Experience.container.tsx
```diff
@@ -0,0 +1,29 @@
...
+  return {
+    qualifications: [],
+    skills: SkillsSelectors.selectSkills(state),
+    organisations: OrganisationsSelectors.selectOrganisations(state),
+  }
+}
+const mapDispatchToProps = (dispatch: Dispatch) => {
+  return {
+    fetchSkillsList: () => {
```
cameronolivier: current pattern is to prefix callbacks/handlers with 'on' - and to be linked to the component, not what it's doing - currently this is `onExperienceInit` - or something like that - as the callback name, and we can then dispatch both skills and organisations from the single callback.

cameronolivier: I actually think we should be loading the skills and organisations on app hydrate already - so not sure how useful it would be to load it again here on component init.

cameronolivier: if you keep it here- please make sure it's not doing an api call everytime we navigate to this page. I don't think the data will be changing often - and it's a better user experience if the data is already there - than if the user now has to wait for the orgs or skills list to populate.

src/modules/Experience/Experience.tsx
```diff
@@ -1,18 +1,24 @@
...
 import styles from './Experience.styles'
-import { ExperienceType } from './Experience.types'
+import { ExperienceFormState, ExperienceType } from './Experience.types'
 import ExperienceForm from './ExperienceForm/ExperienceForm'
 
 interface Props {
+  onExperienceSave: (qualification: any) => void
+  fetchOrganizationsList: () => void
+  fetchSkillsList: () => void
+  qualifications: [QualificationRequestPayload]
```
cameronolivier: is this correct or should it be `QualificationRequestPayload[]`

src/modules/Experience/Experience.tsx
```diff
@@ -26,40 +32,63 @@ const renderItem = ({ job, startDate, endDate }: ExperienceType) => (
   />
 )
 
-const Experience = ({ navigation }: Props) => {
+const Experience = ({
+  navigation,
+  onExperienceSave,
+  fetchOrganizationsList,
```
cameronolivier: as above-  these should be a single callback: something like `onExperienceInit` - that then dispatches for both of these (presuming that's still something we want to do) - I'd argue it would be better to fetch these on app hydrate.

cameronolivier: Additionally: we should call onExperienceInit in Experience, not from within the ExperienceForm - and do the data fetching from the parent component and simply pass organisations and skills down into the form - rather than passing multiple callbacks and data between components - it's unnecessary imo.

src/modules/Experience/ExperienceForm/ExperienceForm.tsx
```diff
@@ -1,70 +1,62 @@
...
 import { INITIAL_VALUES } from './ExperienceForm.constants'
 import styles from './ExperienceForm.styles'
-import { DropDownOrg } from './ExperienceForm.types'
-import { getOrganizationsList, getSkillsList, submitForm } from './ExperienceForm.utils'
+import { DropDownOrg, DropDownSkill } from './ExperienceForm.types'
 import { ValidationSchema } from './ValidationSchema'
 
 interface Props {
-  navigation: StackNavigationProp<HomeNavigatorParamsList, HomeNavigationRoutes.Experience>
+  fetchOrganizationsList: () => void
```
cameronolivier: as mentioned above - this component should just expect the skills and organisations data to be available to it. - it shouldn't need  handle fetching the data as well.

This will also clean up your component as you won't need to handle setting this data - as it should simply be available.

src/modules/Experience/ExperienceForm/ExperienceForm.types.ts
```diff
@@ -2,6 +2,10 @@ export type DropDownOrg = {
   label: string
   value: string
 }
+export type DropDownSkill = {
```
cameronolivier: we're just repeating the same thing here - if all the DropDown types are the same, then we should have a single DropDown type instead of individually replicating the same thing each time.

src/modules/Organisations/Organisations.middleware.tsx
```diff
@@ -0,0 +1,55 @@
+import {
+  fetchOrganisations,
+  fetchOrganisationsFailure,
+  fetchOrganisationsSuccess,
+  setOrganisations,
+} from 'modules/Organisations/Organisations.reducer'
```
cameronolivier: this is the same module - please adjust import to be './Organisations.reducer'

src/modules/Organisations/Organisations.middleware.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { mergeRight } from 'ramda'
+import { Middleware } from 'redux'
+import { showSimpleMessage } from 'utils/error'
+
+import { actions as ApiActions } from '../../api'
+import { constants as ApiOrganisationConstants } from '../../api/organisations'
+import { selectOrganisationsFromPayload } from './Organisations.utils'
```
cameronolivier: `extractOrganisationsFromPayload`

src/redux/store.ts
```diff
@@ -10,6 +10,7 @@ const persistConfig = {
   key: 'root',
   version: 1,
   storage: AsyncStorage,
+  // whitelist: ['auth', 'user'],
```
cameronolivier: remove

src/redux/middleware.ts
```diff
@@ -1,12 +1,16 @@
 import { setItemAsync } from 'expo-secure-store'
 import { middleware as appMiddleware } from 'modules/App'
+import Navigation from 'modules/Navigation'
```
cameronolivier: we've injected Navigation directly into our middleware everywhere else- I'm happy with us taking this stance, as I think it's best to remain consistent as much as possible with dependencies, but then we will need to add a refactor task to update our current middleware to match.

src/modules/Skills/Skills.utils.tsx
```diff
@@ -0,0 +1,3 @@
+import { path } from 'ramda'
+
+export const selectSkillsFromPayload = path(['payload', 'data', 'data'])
```
cameronolivier: use `extract` naming convention.

Please remember to add tests.

src/modules/Skills/Skills.types.tsx
```diff
@@ -0,0 +1,5 @@
+export type Skill = {
+  id?: string
```
cameronolivier: are these really optional? surely a Skill will have an id?

src/modules/Skills/Skills.selector.tsx
```diff
@@ -0,0 +1,3 @@
+import { RootState } from 'redux/redux.types'
+
+export const selectSkills = (state: RootState) => state.skills
```
cameronolivier, line 9: remember to add tests

src/modules/Skills/Skills.reducer.tsx
```diff
@@ -0,0 +1,15 @@
...
+const name = '[Skills]'
+export const INITIAL_STATE = []
+export const setSkills = createAction<any>(`${name} setSkills`)
+
+export const fetchSkills = createAction(`${name} fetchSkills`)
+export const fetchSkillsSuccess = createAction<any>(`${name} fetchSkillsSuccess`)
+export const fetchSkillsFailure = createAction<string>(`${name} fetchSkillsFailure`)
+
+const SkillsReducer = createReducer(INITIAL_STATE, builder => {
+  builder.addCase(setSkills, (state, action) => state.concat(action.payload))
```
cameronolivier: remember to add tests.

Also as mentioned in credentials, please normalise the data - so the state should be an object not an array.

src/modules/Skills/Skills.middleware.tsx
```diff
@@ -0,0 +1,50 @@
+import { fetchSkills, fetchSkillsFailure, fetchSkillsSuccess, setSkills } from 'modules/Skills/Skills.reducer'
+import { mergeRight } from 'ramda'
+import { Middleware } from 'redux'
+import { showSimpleMessage } from 'utils/error'
```
cameronolivier: is this the correct import?

src/modules/Skills/Skills.middleware.tsx
```diff
@@ -0,0 +1,50 @@
+import { fetchSkills, fetchSkillsFailure, fetchSkillsSuccess, setSkills } from 'modules/Skills/Skills.reducer'
```
cameronolivier: please adjust import to './Skills.reducer'

src/modules/Organisations/Organisations.utils.tsx
```diff
@@ -0,0 +1,3 @@
+import { path } from 'ramda'
+
+export const selectOrganisationsFromPayload = path(['payload', 'data', 'data'])
```
cameronolivier: `extract...`

and add tests

src/modules/Skills/Skills.middleware.tsx
cameronolivier: remember to add tests

src/modules/Organisations/Organisations.types.ts
```diff
@@ -0,0 +1,19 @@
+export type OrganisationResponsePayload = {
```
cameronolivier: I'd say this is less the response payload - more just an Organisation, hey? (similar for other places - like skills) - I'm presuming this just represents the Organisation - so it makes sense to call it as such - presuming we're going to be using this as the Organisation in the app

src/modules/Organisations/Organisations.reducer.tsx
```diff
@@ -0,0 +1,15 @@
+import { createAction, createReducer } from '@reduxjs/toolkit'
+
+const name = '[Organisations]'
+export const INITIAL_STATE = []
```
cameronolivier: as other places this should be an object and the data should be normalised.

src/modules/Organisations/Organisations.reducer.tsx
```diff
@@ -0,0 +1,15 @@
...
+export const setOrganisations = createAction<any>(`${name} setOrganisations`)
+
+export const fetchOrganisations = createAction(`${name} fetchOrganisations`)
+export const fetchOrganisationsSuccess = createAction<any>(`${name} fetchOrganisationsSuccess`)
+export const fetchOrganisationsFailure = createAction<string>(`${name} fetchOrganisationsFailure`)
+
+const OrganisationsReducer = createReducer(INITIAL_STATE, builder => {
+  builder.addCase(setOrganisations, (_, action) => action.payload)
```
cameronolivier: reminder to add tests

src/modules/Organisations/Organisations.middleware.tsx
```diff
@@ -0,0 +1,55 @@
+import {
```
cameronolivier: reminder to add tests

src/modules/Experience/Experience.selector.tsx
cameronolivier: the default selector should collate all the data required by the Experience component -in this case it includes Skills and orgs - so add those here - not in the container.

src/modules/Credentials/Credentials.middleware.test.tsx
cameronolivier: obviously need more tests here.

cameronolivier (changes requested):

src/components/DropDownTags/DropDownTags.tsx
```diff
@@ -26,10 +26,6 @@ const DropDownTags = ({ name, label, handlers, ...props }: Props) => {
   const { handleChange, handleBlur, errors, values, touched, setFieldValue } = handlers
   const { t } = useTranslation()
 
-  useEffect(() => {
-    setDropdownValue(values[name])
-  }, [name, values])
-
   const deleteSkill = (tag: string) => setDropdownValue(dropElement(tag, dropDownValue))
```
cameronolivier: considering this is a reusable component called 'DropDownTags' we should remove reference to 'skill' here - probably rename to 'removeTag'? or something?

src/modules/Credentials/Credentials.reducer.tsx
```diff
@@ -1,25 +1,26 @@
 import { createAction, createReducer } from '@reduxjs/toolkit'
+import { mergeDeepRight } from 'ramda'
 
-import { Credential } from './Credentials.types'
+import { CredentialRequestPayload } from './Credentials.types'
```
cameronolivier: why the renaming? is Credential incorrect here? are we using it elsewhere? I would consider naming it UserCredential though

src/redux/store.ts
```diff
@@ -10,7 +10,7 @@ const persistConfig = {
   key: 'root',
   version: 1,
   storage: AsyncStorage,
-  // whitelist: ['auth', 'user'],
+  whitelist: ['auth', 'user'],
```
cameronolivier: why are we only persisting this data?

src/redux/reducers.ts
```diff
@@ -1,11 +1,17 @@
...
 import { reducer as auth } from '../modules/Auth'
+import { reducer as credentials } from '../modules/Credentials'
+import { reducer as organisations } from '../modules/Organisations'
+import { reducer as skills } from '../modules/Skills'
 import { reducer as user } from '../modules/User'
 
 const rootReducer = combineReducers({
   auth,
   user,
+  credentials,
```
cameronolivier: please remember to add their types to the RootState type and their initial states to the state fixture.

cameronolivier (changes requested):

src/modules/Skills/Skills.selector.tsx
sciphergfx: Sure

src/modules/Skills/Skills.middleware.tsx
sciphergfx: import { showSimpleMessage } from 'utils/error' ?

src/modules/Experience/ExperienceForm/ExperienceForm.constants.ts
```diff
@@ -1,18 +1,15 @@
...
+//TODO: Review default values; language, url
+export const INITIAL_VALUES: QualificationRequestPayload = {
   title: '',
   description: '',
-  id: '',
-  startDate: null,
-  endDate: null,
-  country: '',
-  skillNames: [],
+  url: 'none',
```
sciphergfx: So I'm not quite sure about the url source and language value

src/api/qualifications/qualifications.constants.ts
```diff
@@ -0,0 +1,13 @@
...
+import { ApiClients, ApiMeta, ApiMethods } from '../api.types'
+
+export const QUALIFICATIONS_CONFIG: Partial<ApiMeta> = {
+  client: ApiClients.Qualifications,
+  method: ApiMethods.Post,
+  isTokenRequired: true,
+}
+
+export const QUALIFICATIONS_CREATE_CONFIG: Partial<ApiMeta> = mergeRight(QUALIFICATIONS_CONFIG, {
+  method: ApiMethods.Post,
```
cameronolivier: if this is already the default, you don't really need to restate it here-  can just do this, if there are no overrides:
```ts
export const QUALIFICATIONS_CREATE_CONFIG: Partial<ApiMeta> = QUALIFICATIONS_CONFIG
```

src/modules/App/App.middleware.test.ts
```diff
@@ -1,6 +1,6 @@
 import { createMiddlewareStub } from '../../../tests/tests.utils'
 import { actions as AuthActions } from '../Auth'
-import { actions as UserActions } from '../User'
+import { actions as CredentialsActions } from '../UserCredentials'
```
cameronolivier: I know it's longer. -but could we keep the User prefix : `UserCredentialsActions`.

src/modules/App/App.middleware.ts
```diff
@@ -14,6 +17,9 @@ export const appResetFlow: Middleware =
     if (resetAppData.match(action)) {
       dispatch(AuthActions.clearAuth())
       dispatch(UserActions.clearUser())
+      dispatch(UserActions.clearUser())
```
cameronolivier: duplicate clearUser()

src/modules/App/App.middleware.ts
```diff
@@ -25,7 +31,9 @@ export const hydrateAppFlow: Middleware =
   action => {
     const result = next(action)
     if (hydrateApp.match(action)) {
-      dispatch(UserActions.fetchUserCredentials())
+      dispatch(CredentialsActions.fetchUserCredentials())
```
cameronolivier: as mentioned above - should be `UserCredentialsActions`

src/modules/Experience/Experience.container.tsx
```diff
@@ -0,0 +1,18 @@
+import { Dispatch } from '@reduxjs/toolkit'
+import { actions as QualificationsActions } from 'modules/Qualifications'
+import { QualificationRequestPayload } from 'modules/Qualifications/Qualifications.types'
```
cameronolivier: if you exported  this from the Qualifications index.ts then you could import all from Qualifications in a single import: 
```ts
import { actions as QualificationsActions, types as QualificationsTypes } from 'modules/Qualifications'
```

src/modules/Experience/Experience.selector.tsx
```diff
@@ -0,0 +1,9 @@
...
+import { RootState } from 'redux/redux.types'
+
+export const selectState = (state: RootState) => state
```
cameronolivier: we don't need a selectState selector - state _is_ state...

src/modules/Experience/Experience.selector.tsx
```diff
@@ -0,0 +1,9 @@
...
+export default createSelector(selectState, ({ qualifications, skills, organisations }) => ({
```
cameronolivier: ...so this should just be
```ts
export default createSelector((seletQualifications, selectSkills, selectOrganisations) => ({
...
```
Also, you want to create selectors in each module that handle getting the state objects for that module - so qualifications would have a `selectQualifications` selector, etc.

`createSelector` returns a function that takes state and passes it into each of the `selectors`.

src/modules/Experience/ExperienceForm/ExperienceForm.constants.ts
cameronolivier: we getting this from Gareth?

src/modules/Organisations/Organisations.reducer.tsx
```diff
@@ -0,0 +1,17 @@
...
+export const setOrganisations = createAction<any>(`${name} setOrganisations`)
+
+export const fetchOrganisations = createAction(`${name} fetchOrganisations`)
+export const fetchOrganisationsSuccess = createAction<any>(`${name} fetchOrganisationsSuccess`)
+export const fetchOrganisationsFailure = createAction<string>(`${name} fetchOrganisationsFailure`)
+export const clearOrganisations = createAction(`${name} clearOrganisations`)
+
+const OrganisationsReducer = createReducer(INITIAL_STATE, builder => {
+  builder.addCase(setOrganisations, (_, action) => action.payload)
+  builder.addCase(clearOrganisations, () => INITIAL_STATE)
```
cameronolivier: not sure it makes a difference - but I think setting action and state helped in a different case - `(_state, _action) => INITIAL_STATE`

src/modules/Qualifications/Qualifications.utils.tsx
```diff
@@ -0,0 +1,3 @@
+import { path } from 'ramda'
+
+export const extractQualificationsFromPayload = path(['payload', 'data', 'data'])
```
cameronolivier: we're repeating this pattern quite a lot- if this is the default way we get any data out of our payloads, maybe we need a `normalizePayload` function that just does this? Happy to discuss thoughts further.

src/modules/Skills/Skills.middleware.tsx
cameronolivier: yeah -I've since looked at it - just seems a weirdly named import - but it is correct- I think we need to name it better/ refactor it at some point... but ignore for now

src/modules/User/User.reducer.ts
```diff
@@ -41,7 +37,7 @@ export const updateUserPhotoFailure = createAction<UpdateUserFailureResponse>(`$
 
 const UserReducer = createReducer(INITIAL_STATE, builder => {
   builder.addCase(setUser, (state, action) => mergeDeepRight(state)(action.payload))
-  builder.addCase(clearUser, (_state, _action) => INITIAL_STATE)
+  builder.addCase(clearUser, () => INITIAL_STATE)
```
cameronolivier: remember we had a weird issue last time - let's keep an eye on this.

src/utils/error.utils.ts
```diff
@@ -0,0 +1,3 @@
+import { path } from 'ramda'
+
+export const extractErrorMessageFromPayload = path(['payload', 'data', 'meta', 'message'])
```
cameronolivier: I think this might be better placed in `Error/Error.utils`

cameronolivier (changes requested):
I'm about 50% through this, but have a few changes which I thought I'd send through so long.

src/modules/App/App.middleware.test.ts
sciphergfx: missed. that

src/modules/Experience/ExperienceForm/ExperienceForm.constants.ts
sciphergfx: yes

src/utils/error.utils.ts
sciphergfx: I agree

src/modules/Experience/Experience.selector.tsx
sciphergfx: I see now on the documentation that the function allows for multiple selectors, I'll refactor. Thanks

src/modules/Qualifications/Qualifications.utils.tsx
sciphergfx: Added a TODO here

src/modules/Experience/Experience.tsx
```diff
@@ -1,18 +1,23 @@
...
 
 import styles from './Experience.styles'
-import { ExperienceType } from './Experience.types'
+import { ExperienceFormState, ExperienceType } from './Experience.types'
 import ExperienceForm from './ExperienceForm/ExperienceForm'
+import { DropDownList } from './ExperienceForm/ExperienceForm.types'
 
 interface Props {
+  onExperienceSave: (qualification: QualificationRequestPayload) => void
+  qualifications: []
```
cameronolivier: why do we have the 2 different naming conventions - both experience and qualifications seem to be the same thing -  we should pick one unless there are specific differences - and then stick to them. We shouldn't have a component called Experience, which only has qualifications, on onExperienceSave taking a qualification. It's not ideal.

src/modules/Experience/ExperienceForm/ExperienceForm.tsx
```diff
@@ -73,19 +60,21 @@ const ExperienceForm = forwardRef(({ navigation }: Props, ref) => {
...
             }
           />
-          <Spinner visible={formikHandlers.isSubmitting} />
+          {/* <Spinner visible={formikHandlers.isSubmitting} /> */}
           <Input name={'title'} label={t('Title')} handlers={formikHandlers} />
           <DropDown
-            items={organizations}
-            name={'organisationName'}
+            items={mapToDropDownArray(organisationsList, 'key', 'value')}
+            name={'organisationId'}
```
cameronolivier: no need to wrap a string with {} for component props.

src/modules/Experience/ExperienceForm/ExperienceForm.tsx
```diff
@@ -73,19 +60,21 @@ const ExperienceForm = forwardRef(({ navigation }: Props, ref) => {
               'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec quis mauris purus. Quisque malesuada ornare mauris sed feugiat. Cras lectus est, iaculis quis nulla cursus, finibus gravida massa. Donec condimentum porta nisi, eu egestas risus ullamcorper in. In et magna mauris. '
             }
           />
-          <Spinner visible={formikHandlers.isSubmitting} />
+          {/* <Spinner visible={formikHandlers.isSubmitting} /> */}
           <Input name={'title'} label={t('Title')} handlers={formikHandlers} />
           <DropDown
-            items={organizations}
-            name={'organisationName'}
+            items={mapToDropDownArray(organisationsList, 'key', 'value')}
```
cameronolivier: just thinking- I think we should adjust the data so that we're running `mapToDropDownArray` - I actually think the best place is in the selector.

The idea with a selector is that it takes whatever data is required from the state and manipulates it so that it's ready for the component. 

I think that if we're doing the mapping of the data `mapToDropDownArray` inside the component-  that is running every time the component reloads - and as we're seeing - that's super expensive with skills. 

So there are probably even better ways to make this faster (with an autocomplete as an example), but as the first port of call - let's move the mapping of the data to what's required in the dropdown (As that's the only place we're using that) - in the selector. so that the organization list and skills etc are already ready for the dropdown data.

src/modules/Experience/ExperienceForm/ExperienceForm.tsx
```diff
@@ -96,13 +85,14 @@ const ExperienceForm = forwardRef(({ navigation }: Props, ref) => {
             onPress={() => setIsWorkingHere(!isWorkingHere)}
           />
           <View style={styles.row}>
-            <DatePicker name={'startDate'} label={t('Start date')} handlers={formikHandlers} />
-            <DatePicker name={'endDate'} label={t('End date')} handlers={formikHandlers} />
+            <DatePicker name={'startTime'} label={t('Start date')} handlers={formikHandlers} />
```
cameronolivier: Again props with `prop='string'` or `prop={not string}`.

src/modules/Organisations/Organisations.reducer.tsx
```diff
@@ -0,0 +1,17 @@
...
+const name = '[Organisations]'
+export const INITIAL_STATE = [] as any
+export const setOrganisations = createAction<any>(`${name} setOrganisations`)
+
+export const fetchOrganisations = createAction(`${name} fetchOrganisations`)
+export const fetchOrganisationsSuccess = createAction<any>(`${name} fetchOrganisationsSuccess`)
+export const fetchOrganisationsFailure = createAction<string>(`${name} fetchOrganisationsFailure`)
+export const clearOrganisations = createAction(`${name} clearOrganisations`)
+
+const OrganisationsReducer = createReducer(INITIAL_STATE, builder => {
```
cameronolivier: should not be PascalCase

src/modules/Skills/Skills.reducer.tsx
```diff
@@ -0,0 +1,17 @@
+import { createAction, createReducer } from '@reduxjs/toolkit'
+
+const name = '[Skills]'
+export const INITIAL_STATE = [] as any
```
cameronolivier: should this not be something like `Skill[]`? Also I think it would be a good idea to normalise the state data as:

```ts
skills: {
  items: { 
    id1: { ... },
    id2: { ... },
    ....
  },
 itemIds: ['id1', 'id2', ...]
}
```

src/modules/Skills/Skills.types.tsx
```diff
@@ -0,0 +1 @@
+export type Skills = string[]
```
cameronolivier: if the payload seems to be in the shape of
```ts
[{key: '', value: ''}, ...]
```
where do we use Skills as `string[]`? 

and I think it would be worthwhile to type the skills data in state.

src/modules/Experience/Experience.container.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { connect } from 'react-redux'
+
+import Experience from './Experience'
+import selector from './Experience.selector'
+
+const mapStateToProps = selector
+const mapDispatchToProps = (dispatch: Dispatch) => {
+  return {
+    onExperienceSave: (qualification: QualificationsTypes.QualificationRequestPayload) => {
+      dispatch(QualificationsActions.createQualifications(qualification))
```
cameronolivier: I'm not sure this is doing the correct thing. I'm presuming that adding experience is something we're adding to the user - which means it should be a patch to the user's credentials, right? 

Also  - a user's experience is added by jobs - not qualifications. 

Qualifications are a separate thing - and that is added by a user's Education (self-attested) and Courses (created by admin). 

So it doesn't seem that qualifications should feature at all in experience.

cameronolivier (changes requested):

src/modules/Experience/Experience.container.tsx
cameronolivier: Also - I think we should be adding (patching) to the user's credentials - so experience (or job) as part of the user's credentials - not up to the global job or qualifications etc endpoint. 

So to that - we need to be careful with our naming of our modules and our state keys - because 'qualifications' (or whathaveyou) for the user won't be the same as all qualifications, obviously.

Our options as I can see it are: 
1 - put all in `user` with subproperties to hold each credential, alongside other user data.
```ts
state: {
  user: {
    name: 'Joe Soap',
    ...
    experience: {
      byId: {
        id1: {...},
        ...
      },
      allIds: ['id1', 'id2', ...]
    }
  }
}
```

2 - in a `userCredentials` key, so pretty much the same as above but with the usercredentials split out
```ts
state: {
  user: {
    name: 'Joe Soap',
    ...
  },
  userCredentials: {
    experience: {
      byId: {
        id1: {...},
        ...
      },
      allIds: ['id1', 'id2', ...]
    }
  }
}
```

3 - each credential in its own state key: 
```ts
state: {
  user: {
    name: 'Joe Soap',
    ...
  },
  userExperience: {
    byId: {
      id1: {...},
      ...
    },
    allIds: ['id1', 'id2', ...]
  },
  userSkills: {
    byId: {
      id1: {...},
      ...
    },
    allIds: ['id1', 'id2', ...]
  },
  ...
}
...
```

src/modules/UserCredentials/UserCredentials.middleware.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+export const fetchUserCredentialsSuccessFlow: Middleware =
+  ({ dispatch }) =>
+  next =>
+  action => {
+    const result = next(action)
+
+    if (fetchUserCredentialsSuccess.match(action)) {
+      const userCredentials = extractUserCredentialsFromPayload(action)
+      dispatch(setUserCredentials(userCredentials))
```
cameronolivier: I don't believe we should be setting this as a single big blob - I was envisioning that each category (experience, skills, etc) would have its own endpoint with normalised data.

cameronolivier: I have added another comment above around how to go about structuring the user credential data in state (and by extension in our modules). I believe that should inform how we go about doing this. 

I had mentioned something like this in earlier comments on the pr

src/modules/UserCredentials/UserCredentials.test.fixtures.ts
```diff
@@ -0,0 +1,40 @@
+import { createFixture } from '../../../tests/tests.utils'
+import { INITIAL_STATE } from './UserCredentials.reducer'
+
+export const USER_CREDENTIALS_RESPONSE = [
```
cameronolivier: I don't feel like this is a comprehensive example of the data returned (there's definitely more than just qualifications coming back, surely?

src/modules/UserCredentials/UserCredentials.selector.test.tsx
```diff
@@ -0,0 +1,25 @@
...
+      const result = SUT.selectUserCredentials(stateMock)
+      // then ... should return result as expected
+      expect(result).toEqual(USER_CREDENTIALS_RESPONSE)
+    })
+    it('should return the default user credentials state', () => {
+      const state = rootStateFixture()
+      // when ... we call the selector
+      const result = SUT.selectUserCredentials(state)
+      // then ... should return result as expected
+      expect(result).toEqual(state.userCredentials)
```
cameronolivier: This should rather be `USER_CREDENTIALS_RESPONSE` - it's just a little more explicit.

src/modules/Organisations/Organisations.reducer.tsx
```diff
@@ -0,0 +1,17 @@
+import { createAction, createReducer } from '@reduxjs/toolkit'
+
+const name = '[Organisations]'
+export const INITIAL_STATE = [] as any
```
cameronolivier: as we're going to be using INITIAL_STATE for test fixtures and RootState shape, etc  - this needs to be accurate - we can't really be using any here, so please add typing. 

Also as I've mentioned elsewhere, we need to have a normalised data shape.

cameronolivier (changes requested):

src/modules/Experience/Experience.tsx
sciphergfx: We have that because We're creating modules as per the collections on the API, i think you mentioned this.

src/modules/Experience/Experience.container.tsx
sciphergfx: Thanks @cameronolivier 

I think as discussed in yesterdays meeting the swagger documentation does not align with the shape of request data on the fields presented on the experience form on mobile, hence opting for qualifications which included startTime and endTime fields.

That said, the choice to create a separate module for qualifications (which should be jobs) was to maintain our collections module naming standard which I think you suggested. 
I’e Auth / User / “CREDENTIAL_NAME” / “ANOTHER_CREDENTIAL_NAME” … 

I think most have their own CRUD end points has should ideally have their own reducers, if I’m not mistaken? 
I also think this aligns with our reducer names hence my choice for using “credentials” before refactoring to userCredentials as you once suggested we have - auth, user, skills, ( and possibly “credentials” which maintains a state array of credentials ) etc

I think with regards our state shape suggestions, I do agree with option 2, 

I think given the shape of our credentials response payload.
 // credential types illustrated - challenge and job
[  
    {
      "challenge": {
        "id": "challenge_id",
      }, 
      "id": "credential_id",
    }, 
      {
      "challenge": {
        "id": "challenge_id",
      },
      "id": "credential_id",
    },
     {
      "job": {
        "id": "job_id",
      },
      "id": "credential_id",
    },
  ],

And given this data shape
 experience: {
      byId: {
        id1: {...},
        ...
      },
      allIds: ['id1', 'id2', ...]
    }


My concern is having to maintain an array of ids when we don’t have these objects keyed with IDs, something similar to how Firestore stores it’s collection.

If this approach provides us with a much more efficient way of handling the data in-app and doesn’t add any layer of complexity and boilerplate, happy with it!


As a aside, a suggestion:

I think we can still have a “credentials” module and simply extract all credential by type, with one utility function, then make the data available to the credentials, in this case, “challenge” or “job”.

That said, we have an endpoint for fetching credentials by type, which can update those credential types (challenge / job) states seamlessly as they share the same object type

I.E
 src -> module -> credential -> challenge (also contains reducer/middleware for CRUD ops.)
							-> job (also contains reducer/middleware for CRUD ops.)

This will make it much more easier to handle at the UI level

src/modules/Experience/Experience.tsx
sciphergfx: I agree, as. i mentioned above, was trying to avoid wiring all "qualifications" related stuff to experience directly
. I'll rename onExperienceSave

src/modules/Skills/Skills.types.tsx
sciphergfx: I only focused on skills array, which is what's needed for this ticket, I think there's a separate ticket for skills

src/modules/UserCredentials/UserCredentials.test.fixtures.ts
sciphergfx: You're right, but the plan is to handle this on YOMA-451.. I added a TODO to some parts related to this

src/modules/UserCredentials/UserCredentials.middleware.tsx
sciphergfx: I suggested an approach to handling this above.

src/modules/Experience/ExperienceForm/ExperienceForm.tsx
sciphergfx: I think this is allowed,  We have hundreds of it in the code base, but it's a cleaner way to add a string. Do you want me to effect this change on this PR or we create a ticket an effect it everywhere? @cameronolivier

src/modules/Experience/ExperienceForm/ExperienceForm.tsx
sciphergfx: I think this is allowed,  We have many instances of it around the code base, but it's a cleaner way to add a string. Do you want me to effect this change on this PR or we create a ticket an effect it everywhere? @cameronolivier

src/modules/Experience/ExperienceForm/ExperienceForm.tsx
sciphergfx: Responded to this above

src/modules/Organisations/Organisations.reducer.tsx
sciphergfx: I agree

src/modules/Skills/Skills.reducer.tsx
sciphergfx: this will most likely change when we get to the skills ticket, I added this to hold an array of skills for the experience form

src/modules/Organisations/Organisations.reducer.tsx
sciphergfx: You mean the reducer variable name?

src/modules/Experience/ExperienceForm/ExperienceForm.tsx
cameronolivier: weird. it should be caught by linting. It is the pattern we should be sticking to - but refactor when you touch the file - so if you're making changes now, then please fix it up. I don't think we need a separate PR for this.

src/modules/Experience/ExperienceForm/ExperienceForm.tsx
cameronolivier: @sciphergfx  - have you gotten to this?

src/modules/Organisations/Organisations.reducer.tsx
cameronolivier: yes

src/modules/Skills/Skills.reducer.tsx
cameronolivier: I think the skills ticket is the skills for the user. But happy so long as we ensure we're updating within the next few weeks.

src/redux/store.ts
```diff
@@ -11,6 +12,7 @@ const persistConfig = {
   key: 'root',
   version: 1,
   storage: AsyncStorage,
+  blackLists: ['skills'],
```
sciphergfx: Using skills as a search currently, the data is not persisted, and it doesn't need to be cleared as it's temporarily hence removing it on resetAppState. I'll refactor this when I work on the skills ticket.

src/modules/Job/Job.utils.tsx
```diff
@@ -0,0 +1,15 @@
+import { UserCredentialTypes } from 'api/users/users.types'
+import { path, pick, pipe } from 'ramda'
+
+import { JobCredentialsTmpFormValues, JobPayload } from './Job.types'
+
+//TODO: move to src/utils as normalizePayload
+export const extractJobsFromPayload = path(['payload', 'data', 'data'])
+export const extractJobsCredentialTmpFormValues = pipe(path(['payload']), pick(['startTime', 'endTime']))
+export const prepareJobCredentialPayload =
```
sciphergfx, line 9: can use "extract", not sure if it fits well here

src/modules/Job/Job.selector.tsx
```diff
@@ -0,0 +1,7 @@
+import { createSelector } from '@reduxjs/toolkit'
+import { path } from 'ramda'
+import { RootState } from 'redux/redux.types'
+
+export const selectJob = (state: RootState) => state.job
+export const selectJobTmpFormValues = createSelector(selectJob, path(['tmpFormValues']))
```
cameronolivier: what is this for? why are we storing `tmpFormValues`?

src/components/InfoCard/InfoCard.styles.ts
```diff
@@ -24,6 +24,7 @@ const styles = {
     height: IMAGE_SIZE,
     width: IMAGE_SIZE,
     marginRight: 10,
+    borderRadius: 50,
```
cameronolivier: Why 50? If we want to create a circular radius, it should be 50% of the width - so set it to IMAGE_SIZE/2 if that is what you want to do. 

you can import `CIRCULAR_RADIUS_DIVISOR` to make it more explicit: 
```ts
borderRadius: IMAGE_SIZE/CIRCULAR_RADIUS_DIVISOR,
```

src/utils/strings.utils.ts
```diff
@@ -7,6 +8,17 @@ export const mapToDropDownArray = (array: Record<string, string>[], valueProp =
   }))
 }
 
+export const dropDownFromArray = (array: string[]) => {
```
cameronolivier: this and the above isn't a string util. Maybe an array util... but I think this and the above should move to the dropdown component - as they're specifically for that purpose.

src/utils/redux.utils.ts
```diff
@@ -1,11 +1,14 @@
...
 export const normalise = pipe(
+  ifElse(is(Array), identity, append(__, [])),
   objFromListWith(prop('id')),
   applySpec({
     entities: identity,
     ids: keys,
   }),
 )
+
+export const extractPayloadData = path(['payload', 'data', 'data'])
```
cameronolivier: This wouldn't be a redux util - I'd think it'd be an api util?

src/modules/User/User.types.ts
```diff
@@ -47,3 +47,24 @@ export interface UserResponse {
   createdAt: string
   lastLogin: string
 }
+
+export interface UserCredentialPayload {
```
cameronolivier: is this the 'payload' or is it really just a type for a UserCredential (it seems to be the latter)

src/modules/User/User.reducer.ts
```diff
@@ -39,6 +35,10 @@ export const uploadUserPhotoFailure = createAction<string>(`${name} uploadUserPh
 export const updateUserPhotoSuccess = createAction(`${name} updateUserPhotoSuccess`)
 export const updateUserPhotoFailure = createAction<UpdateUserFailureResponse>(`${name} updateUserPhotoFailure`)
 
+export const fetchUserCredentials = createAction(`${name} fetchUserCredentials`)
+export const fetchUserCredentialsSuccess = createAction<any>(`${name} fetchUserCredentialsSuccess`)
```
cameronolivier: Please add a TODO to type this correctly.

src/modules/Skills/Skills.reducer.tsx
```diff
@@ -0,0 +1,32 @@
...
+const name = '[Skills]'
+export const INITIAL_STATE = {
+  filteredSkills: [],
+  skillEntities: [],
+} as SkillsState
+
+export const setSkillEntities = createAction<Skills[]>(`${name} setSkillEntities`)
+export const setFilteredSkills = createAction<Skills[]>(`${name} setFilteredSkills`)
+
+export const filterSkillsByName = createAction<string>(`${name} filterSkillsByName`)
```
cameronolivier: Do we have Name? - Skills Type is `{ key, value }` -> would it be `byValue`?

src/utils/strings.utils.ts
```diff
@@ -7,6 +8,17 @@ export const mapToDropDownArray = (array: Record<string, string>[], valueProp =
...
+  return array.map((opt: string) => ({
+    label: opt,
+    value: opt,
+  }))
+}
+
+export const sliceArrayByNumber = (number: Number, array: any[]) => slice(0, number, array)
```
cameronolivier: this is maybe an array util? not a string one. And would probably be splitAtPosition - or splitAt? not byNumber. 

This ramda util does it already though: https://ramdajs.com/docs/#splitAt

src/utils/strings.utils.ts
```diff
@@ -7,6 +8,17 @@ export const mapToDropDownArray = (array: Record<string, string>[], valueProp =
...
+export const searchArrayOfObjByValue = (query: string, array: any[]) =>
```
cameronolivier: the array shouldn't be of any[] - it should be typed to be an object with at least a value property. 
- at which point, is it useful to be generic? 

if we're simply finding and returning values in an array - I think we can add this to Skills as a util - and simply call it something like findSkills. Also, I'm not sure we need to be checking query length here.

cameronolivier (changes requested):
This should be deleted: `src/modules/Organisations/Organisations.types.tsx` -if it's not already

My expectation of this task was to simply handle saving new or edited experience to the backend. That's it 

you should have split out the other requirements into separate stories or subtasks if needed. 

Handling the data for the UI to enable being able to send should be stories that come before this - that we can PR separately. 

Could you break it up a bit better. I can't review 71 files effectively.

src/utils/redux.utils.ts
sciphergfx: sure

src/modules/Experience/Experience.utils.tsx
```diff
@@ -0,0 +1,14 @@
+import { applySpec, path, pathOr, prop } from 'ramda'
+
+export const extractExperienceFormValues = applySpec({
+  id: path(['job', 'id']),
+  credentialId: path(['id']),
+  title: path(['job', 'title']),
+  description: path(['job', 'description']),
+  language: path(['job', 'language']),
+  published: path(['job', 'published']),
+  skillNames: pathOr([], ['job', 'skills']),
```
sciphergfx: Api doesn't include skills array hence using pathOr

src/modules/User/User.types.ts
sciphergfx: there are 2 payloads for userCredentials create/update hence seperating this

src/modules/Job/Job.selector.tsx
sciphergfx: We need a way to temporarily store form values needed for credential creation and updates.

src/utils/strings.utils.ts
sciphergfx: I'll take this out, it was added when I implemented a search API for skills

src/components/DropDownTags/DropDownTags.tsx
```diff
@@ -45,9 +47,11 @@ const DropDownTags = ({ name, label, handlers, ...props }: Props) => {
         searchContainerStyle={styles.searchContainer}
         listMode={'MODAL'}
         onChangeValue={itemValue => {
-          handleChange(name)
-          handleBlur(name)
-          setFieldValue(name, itemValue)
+          if (values[name] !== itemValue) {
```
cameronolivier, line 50: Suggestion: I think it might be better to move this to a callback function? I"m not sure how heavy these functions might be? but we're calling 3 and reinstantiating this function on every refresh. What do you think?

src/modules/UserJobs/UserJobsForm/UserJobsForm.types.ts
```diff
@@ -0,0 +1,4 @@
+export type DropDownList = {
```
cameronolivier: Question: is this not a type shared across multiple components? Surely we have a shared type for this already? 
Suggestion: If not we should create one.

src/modules/UserJobs/UserJobs.tsx
```diff
@@ -0,0 +1,104 @@
...
+  }
+  const handleUserJobsFormSave = () => {
+    if (!isEditMode) {
+      onJobCreate(formState.values)
+    } else {
+      onJobPatch(formState.values)
+    }
+  }
+
+  const renderItem = (item: any) => {
```
cameronolivier: issue: could you type this

src/modules/UserJobs/UserJobs.tsx
```diff
@@ -0,0 +1,104 @@
...
+import { HomeNavigationRoutes, HomeNavigatorParamsList } from 'modules/HomeNavigation/HomeNavigation.types'
+import React, { useState } from 'react'
+import { useTranslation } from 'react-i18next'
+import { FlatList, ScrollView } from 'react-native'
+
+import styles from './UserJobs.styles'
+import * as UserJobsTypes from './UserJobs.types'
+import { extractUserJobsFormValues } from './UserJobs.utils'
+import UserJobsForm from './UserJobsForm/UserJobsForm'
+import { INITIAL_VALUES, MOCK_ORGANISATIONS, MOCK_SKILLS } from './UserJobsForm/UserJobsForm.constants'
```
cameronolivier: issue: we should remove the mock values as by this task (452) everything should be tied up and we should no longer need them. Can delete out of the constants file as well.

src/modules/UserJobs/UserJobs.tsx
```diff
@@ -0,0 +1,104 @@
...
+  const [isSaved, setIsSaved] = useState(false)
+  const [isEditMode, setIsEditMode] = useState(false)
+  const [userJobsList] = useState<[]>([])
+  const [formState, setFormState] = useState<UserJobsTypes.UserJobsFormState>({ isValid: true, values: INITIAL_VALUES })
+
+  const addUserJob = () => {
+    setIsSaved(true)
+    setIsEditMode(false)
+  }
+  const editUserJob = (item: any) => {
```
cameronolivier: suggestion: I think it might be good to wrap this in a useCallback.

src/modules/UserJobs/UserJobs.tsx
```diff
@@ -0,0 +1,104 @@
...
+          </Optional>
+        }
+      >
+        <ScrollView>
+          <Card>
+            <UserJobsForm
+              formValues={formState?.values}
+              filterSkillsByValue={filterSkillsByValue}
+              setFormState={setFormState}
+              skills={MOCK_SKILLS}
```
cameronolivier: we should have skills and orgs wired up to real data as a part of this task.

src/modules/UserJobs/UserJobsForm/UserJobsForm.tsx
```diff
@@ -0,0 +1,96 @@
...
+            visible={showInfoModal}
+            closeModal={() => setShowInfoModal(false)}
+            infoText={
+              'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec quis mauris purus. Quisque malesuada ornare mauris sed feugiat. Cras lectus est, iaculis quis nulla cursus, finibus gravida massa. Donec condimentum porta nisi, eu egestas risus ullamcorper in. In et magna mauris. '
+            }
+          />
+          {/* <Spinner visible={formikHandlers.isSubmitting} /> */}
+          <Input name={'title'} label={t('Title')} handlers={formikHandlers} />
+          <DropDown
+            items={mapToDropDownArray(organisationsList, 'key', 'value')}
```
cameronolivier: issue: this mapping for dropdown needs to be happening in the selector, not here.

src/modules/UserJobs/UserJobsForm/UserJobsForm.tsx
```diff
@@ -0,0 +1,96 @@
...
+            label={t('I currently work here')}
+            onPress={() => setIsWorkingHere(!isWorkingHere)}
+          />
+          <View style={styles.row}>
+            <DatePicker name="startTime" label={t('Start date')} handlers={formikHandlers} />
+            <DatePicker name="endTime" label={t('End date')} handlers={formikHandlers} />
+          </View>
+          <Input name={'description'} label={t('Description')} handlers={formikHandlers} multiline />
+          <DropDownTags
+            items={mapToDropDownArray(skillsList, 'value', 'value')}
```
cameronolivier: issue: this mapping for dropdown needs to be happening in the selector, not here.

cameronolivier: this will be re-mapped every time we load this component.

cameronolivier (changes requested):
suggestion: Remember to bump version and release notes.

src/components/DropDownTags/DropDownTags.tsx
sciphergfx: the component re-renders when a change is effected, this was a way to ensure it does so when it's values are changed.  It was a fix to the UI issue with dropdown

src/modules/UserJobs/UserJobs.tsx
sciphergfx: We need the 2 pending PRs in before this change

src/modules/UserJobs/UserJobs.tsx
sciphergfx: replied this above

src/modules/UserJobs/UserJobsForm/UserJobsForm.tsx
sciphergfx: already in

src/modules/Skills/Skills.utils.tsx
```diff
@@ -6,7 +6,7 @@ import { SkillsState } from './Skills.types'
 export const selectFilterSkillsByTerm = curry(({ searchTerm = '', entities, ids }: SkillsState) => {
   if (searchTerm.length < SEARCH_TERM_MIN_LENGTH) {
     const defaultIds = slice(0, FALLBACK_SKILLS_LENGTH, ids)
-    return map((id: string | number) => entities[id].value, defaultIds)
+    return map((id: string | number) => entities[id], defaultIds)
```
cameronolivier: question: why would this be a number? We cannot use numbers as an object key, so I don't think this would every be a number. (if it is we should prefix the number with a string)

src/modules/UserJobs/UserJobs.selector.tsx
```diff
@@ -1,6 +1,24 @@
...
+
+export const selectUserJobItems = createSelector(
+  selectUserJobs,
+  pipe(prop('entities'), values, map(pick(['job', 'startDate', 'endDate']))),
+)
+
+export default createSelector(
+  selectUserJobItems,
+  selectOrganisationsList,
+  selectFiltered,
```
cameronolivier: issue: `selectFiltered` doesn't inform that this is for skills. 
We've chosen to keep relatively generic selector names (`selectId` or `selectFiltered`) - but to make this useful, we need to import external selectors as `import * as`...  - ie `import * as OrganisationsSelectors`  & `import * as SkillsSelectors` so that we can use it as `SkillsSelectors.selectFiltered`

src/modules/Skills/Skills.utils.tsx
cameronolivier: Question: If the shape of entities is `{id: {id: 'id', value: 'value'}}`, why are you returning the entity object?

cameronolivier (changes requested):
a couple of questions and 1 or 2 final updates.

src/modules/Skills/Skills.utils.tsx
sciphergfx: The UI component takes an object hence selector having ```mapToDropDownArray(list, 'value', 'value')```

src/modules/UserJobs/UserJobs.selector.tsx
sciphergfx: Noted

src/utils/redux.utils.ts
```diff
@@ -23,3 +23,4 @@ export const updateNormalisedState = (normalisedState: Record<string, any>, norm
 export const selectNormalised = pick(['ids', 'entities'])
 export const extractId = path(['payload', 'id'])
 export const extractDataFromPayload = path(['payload', 'data', 'data'])
+export const extractUserCredentialId = path(['payload', 'data', 'data', 'id'])
```
cameronolivier: suggestion/nitpick: this should be in User module utils, a module specific util (...UserCredential...) should not be in the shared utils.

However you could rename to `extractId` which you already have. 

I'd suggest you use `extractDataFromPayload` and simply get the id from the 
```ts
const data = extractDataFromPayload(action)
return data.id //or however you want to deal with it.
```
I don't think it needs be any more complicated than that.

src/redux/store.ts
```diff
@@ -27,6 +29,10 @@ const store = configureStore({
           ApiActions.apiRequest.type,
           ApiActions.apiError.type,
           UserActions.uploadUserPhotoSuccess.type,
+          JobsActions.createJob.type,
```
cameronolivier: question: what of these require to be excluded? I'm not sure I understand why.

src/redux/store.ts
```diff
@@ -27,6 +29,10 @@ const store = configureStore({
...
+          UserJobsActions.createUserJobSuccess.type,
+          UserJobsActions.setUserJobsFormValues.type,
```
cameronolivier: issue: we should definitely not be ignoring any `set` actions - we should work around these so that we aren't storing unserialisable data in state - that should never happen. 

The only reason I can see why an error is thrown is that state should only have serialisable data. So currently the app tests all actions - and errors if one doesn't have serialisable data - but for the most part it's fine for us because we're never storing that in state - it's being used within our flows.

So you need to find a way around this without adding this action (or any action tied to a reducer) here.

src/modules/UserJobs/UserJobs.utils.tsx
```diff
@@ -0,0 +1,29 @@
...
+  title: path(['job', 'title']),
+  description: path(['job', 'description']),
+  skillNames: pathOr([], ['job', 'skills']),
+  organisationId: path(['job', 'organisationId']),
+  startTime: prop('startDate'),
+  endTime: prop('endDate'),
+})
+
+export const extractUserJobsFromPayload = pipe(
+  path(['payload', 'data', 'data']),
```
cameronolivier: 1. issue: we already have a util that deals with getting data out of a payload
2. nitpick: 'Job' in `extractUserJobsFromPayload` should be singular if we're only getting 1 job.
3. issue/nitpick: aim to do 1 thing in a function. This function is meant to extract data - don't also cast as an array.

suggestion: 
This util should only care about extracting data out of the data object itself - so we should prepare the data first (extract from payload with `extractDataFromPayload` in `src/util/redux.utils`) and _then_ get what we need from it. Finally, we can cast to array further on in our code - either by calling `of` on it or simply by wrapping it in `[]`.

Rename to: `extractUserJobFromData` (singular 'Job' and extract from data not payload)

```ts
// this util: 
export const extractUserJobFromData = pick([
  'id',
  'verifiedAt',
   ... // all the other properties
  'job',
]),
...

// usage:
const data = extractDataFromPayload(action) // prepare the date by extracting it from the payload
const userJob = extractUserJobFromData(data) // only get the user job from data
const return [userJob] // cast as array
```

src/modules/UserJobs/UserJobs.tsx
```diff
@@ -0,0 +1,120 @@
...
+import UserJobsForm from './UserJobsForm/UserJobsForm'
+import { INITIAL_VALUES } from './UserJobsForm/UserJobsForm.constants'
+
+interface Props {
+  userJobs: UserJobsTypes.UserJobItem[]
+  organisations: GeneralTypes.DropDownList[]
+  skills: GeneralTypes.DropDownList[]
+  onJobCreate: (job: any) => void
+  onJobPatch: (job: any) => void
+  filterSkillsByValue: (query: string) => void
```
cameronolivier: nitpick: this should be named according to convention
suggestion: something like `onFilterSkills`.

src/modules/UserJobs/UserJobsForm/UserJobsForm.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { TouchableOpacity, View } from 'react-native'
+import { Colors } from 'styles'
+import { DropDownList } from 'types/general.types'
+
+import { UserJobFormFields, UserJobsFormState } from '../UserJobs.types'
+import styles from './UserJobsForm.styles'
+import { ValidationSchema } from './ValidationSchema'
+
+interface Props {
+  filterSkillsByValue: (value: string) => void
```
cameronolivier: nitpick: this should be named according to convention.
suggestion: something like `onFilterSkills`.

src/modules/UserJobs/UserJobs.tsx
```diff
@@ -0,0 +1,120 @@
...
+  onJobPatch,
+  filterSkillsByValue,
+  navigation,
+}: Props) => {
+  const { t } = useTranslation()
+  const [isSaved, setIsSaved] = useState(false)
+  const [isEditMode, setIsEditMode] = useState(false)
+  const [formState, setFormState] = useState<UserJobsTypes.UserJobsFormState>({ isValid: true, values: INITIAL_VALUES })
+
+  const addUserJob = useCallback(() => {
```
cameronolivier: nitpick: should follow convention - `handleAddUserJob`

src/modules/UserJobs/UserJobs.tsx
```diff
@@ -0,0 +1,120 @@
...
+    setIsSaved(true)
+    setIsEditMode(false)
+  }, [])
+
+  const editUserJob = useCallback(
```
cameronolivier: nitpick: should follow convention - `handleEditUserJob`

src/modules/UserJobs/UserJobs.middleware.ts
```diff
@@ -38,7 +42,10 @@ export const getUserJobsFromCredentialsFlow =
   next =>
   action => {
     const result = next(action)
-    if (UserActions.fetchUserCredentialsSuccess.match(action)) {
+    if (
+      UserActions.fetchUserCredentialsSuccess.match(action) ||
+      UserActions.fetchUserCredentialsFailure.match(action)
```
cameronolivier: issue/question: why are we catching failure action here and handling it as if it was a success?

src/modules/UserJobs/UserJobs.middleware.ts
```diff
@@ -110,20 +117,69 @@ export const createUserJobFlow: Middleware =
...
+    return result
+  }
+export const fetchUserJobByIdSuccessFlow =
+  ({ notification }: { notification: typeof showSimpleMessage }): Middleware =>
+  ({ dispatch }) =>
+  next =>
+  action => {
+    const result = next(action)
+    if (fetchUserJobByIdSuccess.match(action)) {
+      const response = extractUserJobsFromPayload(action)
```
cameronolivier: suggestion: as mentioned around the utils, this should be adjusted to a 2-step process, similar to line 49 of this file (above).

```ts
const data = extractDataFromPayload(action)
const job = extractUserJob(data)
const normalisedJobs = normalise([job])
```

cameronolivier (changes requested):

src/modules/UserJobs/UserJobs.middleware.ts
sciphergfx: ah....this should have been a reverted error, missed it sorry

src/redux/store.ts
sciphergfx: I understand this could be better handled but it really doesn't matter since the date string is processed nonetheless. I'll try to check the documentation to see how it can be better approached. As an aside, decided to add this owing to time constraint to get this in.

src/redux/store.ts
sciphergfx: responded above

src/redux/store.ts
cameronolivier: My concern is that it shouldn't be throwing an error if it is only serializable data and we shouldn't be saving anything that isn't serializable - that shouldn't go into state. I doubt that 'createJob' is storing to state - so that's fine it's more the `setUserJobsFormValues` that I think we need to remove (or find how to fix).

src/modules/Skills/Skills.reducer.tsx
```diff
@@ -15,7 +15,7 @@ export const getSkillsSuccess = createAction<Skill[]>(`${name} getSkillsSuccess`
 export const normaliseSkillsSuccess = createAction<NormalisedSkills>(`${name} normaliseSkillsSuccess`)
 export const setFilterSearchTerm = createAction<string>(`${name} setFilterSearchTerm`)
 
-export const filterSkillsByValue = createAction<string>(`${name} filterSkillsByValue`)
+export const onFilterSkills = createAction<string>(`${name} onFilterSkills`)
```
cameronolivier: sorry if there was a misunderstanding- the action shouldn't be called 'onFilterSkills' - just the callbacks. 
an action can be `filterSkills` or `getFilteredSkills` (by value would be implied, I'd think) -

src/redux/store.ts
cameronolivier: so did I :) - we should make 100% sure that `setUserJobsFormValues` isn't storing anything unserialisable in state and remove it from this ignore list. 

I'm ok with the others for now (but don't want to be just adding things here unless necessary)

cameronolivier (changes requested):
the `setUserJobsFormValues` is an important thing to fix imo.

cameronolivier (approved):

PR #101, opened 2021-07-22
YOMA-438 - proposed fixes for tests

cameronolivier:
## Work Done
- refactored the utils and middleware around photo upload so that we're passing in all third party dependencies, only testing the upload util once, and passed in the functions for easier mocking in our tests.
- moved the Profile.constants to Users.constants.

## Steps to Test
- run tests - all should pass.
- upload an image


sciphergfx (approved):
All good

PR #103, opened 2021-07-26
YOMA-495 - feat(CvCard): adjust to handle children

cameronolivier:
## Scope [YOMA-495](https://yomaworld.atlassian.net/browse/YOMA-495)
- we need to adjust Cv Card to be able to take children.

## Work Done
- added children to CvCard
- cleaned up the component somewhat, and pulled out some dependency components
- added a `rightAccessory` prop to `ListCardHeader` so we can pass the Edit button directly into the card header.

## Steps to Test
- navigate to MyCV / DigitalCv and make sure it looks ok
- ensure edit buttons work and text etc looks correct
- also make sure that the My Skills view looks fine (the other place that `ListCardHeader` is being used.


src/components/CvCard/CvCard.tsx
```diff
@@ -1,57 +1,38 @@
...
-      <View style={styles.dividerLine} />
-      <View style={styles.bodyView}>{renderContent(content, defaultText)}</View>
+}>
+const CvCard = ({ title, fallback, count, badgeColor = Colors.white, onEdit, children }: Props) => (
+  <View style={styles.container}>
+    <ListCardHeader
+      color={badgeColor}
+      count={count || 0}
+      header={title}
+      hasCountBadge={isNotNil(count) as boolean}
```
sciphergfx, line 26: Why don't we rather have a CvCard.util.tsx that imports a function to handles this, and remove the newly created src/utils/ramda.ts? Ramda is a utility library itself. Your thoughts? @cameronolivier

src/components/ListCardHeader/ListCardHeader.tsx
```diff
@@ -11,18 +11,24 @@ type Props = {
   count: number
   color: Colors
   header: string
+  hasBorder?: boolean
   hasCountBadge?: boolean
+  rightAccessory?: ReactNode
```
sciphergfx: I'm guessing rightAccessory is used to describe the rightComponent? I think the naming can be better

src/vendor.d.ts
```diff
@@ -16,7 +16,7 @@ declare module 'ramda' {
   export function head(...rest: any[]): any
   export function identity(...rest: any[]): any
   export function ifElse(...rest: any[]): any
-  export function isNil(...rest: any[]): any[]
+  export function isNil(...rest: any[]): any | boolean
```
sciphergfx: If we decide to go with a util file for the above, I don't think this would be necessary. I think this introduces a boolean option to the isNil function which does not align with the Ramda documentation if i'm not mistaken?

src/components/CvCard/CvCard.tsx
cameronolivier: Because this is a utility that is re-usable that simply extends ramda's utility functions. So for that I think we can keep the ramda.utils. isNotNil is usually a util fn I end up creating in most of my projects - generally find a need for it - but I think it's worth having things like ramda.utils for these types of cases (or strings.utils for reusable string functions).

We could pull this into a utils file - it's a pretty straightfoward little function though - so not sure if there will be any wins doing it that way, tbh.

src/components/ListCardHeader/ListCardHeader.tsx
cameronolivier: It is a wrapper for something one might want to put on the right of the header - in this case a callback button. I would say we could actually pull out the left badge and add a leftAccessory which simply links that in. 

However, that's probably a step too 'dry' at this point. We don't yet have a usecase for a header component that needs these areas to be flexible - so with that in mind, I'd concur - I'm happy to rename to something more case-specific. 

I'll think about it a bit more - do you have any suggestions?

src/vendor.d.ts
cameronolivier: isNil is always boolean. Or it's a function. this fits it perfectly - I could arguably remove it and simply go with `any` as we have done for all the others - was trying to type this a little closer to its actual use-case. Not sure I've succeeded (`any | boolean` is pretty much the same as `any` anyway), so will adjust back to simply `any`

src/components/ListCardHeader/ListCardHeader.tsx
sciphergfx: Understandable, wasn't thinking maybe "rightComponent' / "rightActionComponent".

src/vendor.d.ts
sciphergfx: shot

src/components/ListCardHeader/ListCardHeader.tsx
cameronolivier: rightComponent is good I think

sciphergfx (dismissed):
All good, app runs!

sciphergfx (approved):
Approved

PR #104, opened 2021-07-31
YOMA-496_setting_up_credential_and_normalising_utils

cameronolivier:
## Scope [YOMA-496](https://yomaworld.atlassian.net/browse/YOMA-496)
 - add helper utils for dta normalising and extrating credentials from the user credentials response

## Work Done
- adding `objFromListWith` as a ramda util which creates an object from a list
- adding `normalise` util to normalise a data set
- adding `extractCredentialsByType` util in User to extract all credentials of a given type from response data

## Steps to Test
run tests


sciphergfx (approved):
All good

PR #105, opened 2021-08-06
YOMA-499 - feat(Organisations) - add organisations data to state

sciphergfx:
## Work Done
- added organisations data to state
- handles fetching organisations only for user experience form


cameronolivier (commented):
Please delete: `src/modules/Organisations/Organisations.types.tsx` as there is nothing in it.

src/modules/Organisations/Organisations.utils.tsx
```diff
@@ -0,0 +1,3 @@
+import { path } from 'ramda'
+
+export const extractOrganisationsFromPayload = path(['payload', 'data', 'data'])
```
cameronolivier: I think I said this elsewhere - this (`['payload', 'data', 'data']`) is a repeated pattern - I think it's best we move it to a single util - api utils might be a good place. 
something like `extractPayloadData` should work.

src/modules/Organisations/Organisations.reducer.tsx
```diff
@@ -0,0 +1,17 @@
+import { createAction, createReducer } from '@reduxjs/toolkit'
+
+const name = '[Organisations]'
+export const INITIAL_STATE = [] as any
+export const setOrganisations = createAction<any>(`${name} setOrganisations`)
```
cameronolivier: please type this correctly.

src/modules/Organisations/Organisations.reducer.tsx
```diff
@@ -0,0 +1,17 @@
+import { createAction, createReducer } from '@reduxjs/toolkit'
+
+const name = '[Organisations]'
+export const INITIAL_STATE = [] as any
```
cameronolivier: please type this correctly

cameronolivier: Also please normalise the data as we have discussed.

src/modules/Organisations/Organisations.reducer.test.tsx
```diff
@@ -0,0 +1,41 @@
+import { rootStateFixture } from 'redux/redux.test.fixtures'
+
+import SUT, { clearOrganisations, INITIAL_STATE, setOrganisations } from './Organisations.reducer'
+
+describe('modules/Organisations/Organisations.reducer', () => {
+  describe('setOrganisations', () => {
+    it('should set organisations correctly', () => {
+      // given ....
+      const mockState = rootStateFixture({})
```
cameronolivier: you don't need to pass an empty object - it's set as the default already.

cameronolivier: call it like `rootStateFixture()`

src/modules/Experience/Experience.selector.test.tsx
```diff
@@ -0,0 +1,18 @@
+import { rootStateFixture } from 'redux/redux.test.fixtures'
+
+import * as SUT from './Experience.selector'
+
+describe('modules/Experience/Experience.selector', () => {
+  describe('selector', () => {
+    it('should return organisations from the root state', () => {
```
cameronolivier: please mock what happens with data as well.

src/modules/Organisations/Organisations.middleware.tsx
```diff
@@ -0,0 +1,58 @@
...
+            onSuccess: fetchOrganisationsSuccess,
+            onFailure: fetchOrganisationsFailure,
+          }),
+        ),
+      )
+    }
+    return result
+  }
+
+export const fetchOrganisationsSuccessFlow: Middleware =
```
cameronolivier, line 37: please add normalising data here as well

src/modules/Organisations/Organisations.middleware.test.tsx
```diff
@@ -0,0 +1,117 @@
...
+          success: true,
+          code: 200,
+          message: null,
+        },
+      }
+
+      const action = fetchOrganisationsSuccess(mockResponseData)
+      // @ts-ignore
+      const { store, invoke } = create(SUT.fetchOrganisationsSuccessFlow)
+      // when ... we respond to the fetchOrganisationsSuccess action
```
cameronolivier, line 97: please add a line of whitespace above the //when.. and //then...

cameronolivier (changes requested):

src/modules/Organisations/Organisations.reducer.tsx
cameronolivier: Normalising is useful as we both want to list all - and handle selecting specific organisations - which is what normalising makes easy and trivial. 

Anywhere where we may want to both map over a list of objects AND need to access specific objects directly which is a powerful paradigm and makes working with the data a lot easier - so wherever we're handling lists of objects let's do this, please.

sciphergfx: This approach was taken as we needed just a list of organisations for the specific ticket we were working on before the request to break the PR down further. If normalising its state data adds any layer of flexibility and speed to its current implementation/use at the UI level, happy to do it.

In order to avoid any back and forth, is the expectations here to use Object.values(theNormalisedEntity) in the UI?

src/modules/Organisations/Organisations.middleware.tsx
sciphergfx: Noted, happy to implement, i've asked a question in this regard, please see below. Thanks

src/modules/Organisations/Organisations.utils.tsx
sciphergfx: Yes added util function in another PR, I'll copy here again.

sciphergfx: Yes, added this util and used in one of the PR it was extracted from. I'll copy over

src/api/api.types.ts
```diff
@@ -17,7 +17,6 @@ export type ApiEndpoints =
 export enum ApiClients {
   Auth = 'auth',
   Users = 'users',
-  Organisations = 'organisations',
```
cameronolivier: issue: this should not be removed.

src/modules/Organisations/Organisations.middleware.tsx
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+export const normaliseOrganisationsFlow =
+  (normalise: StdFn<Organisation[], NormalisedOrganisations>): Middleware =>
+  ({ dispatch }) =>
+  next =>
+  action => {
+    const result = next(action)
+    if (getOrganisationsSuccess.match(action)) {
+      const normalisedOrganisations = normalise(action.payload)
```
cameronolivier: as the data shape is similar to Skills, we will need to wait until the skills branch is in before merging, otherwise running normalise will break (there's no 'id' prop).

cameronolivier (changes requested):
Please add release notes and bump the version.

src/api/api.types.ts
sciphergfx: shot, missed. Will re-add then release notes and version bump

android/app/build.gradle
```diff
@@ -140,7 +140,7 @@ android {
         applicationId "com.yomamobile"
         minSdkVersion rootProject.ext.minSdkVersion
         targetSdkVersion rootProject.ext.targetSdkVersion
-        versionCode 19
+        versionCode 20
```
cameronolivier: nitpick: (blocking) this should be bumped to 22, please. Develop is currently on 21. https://github.com/didx-xyz/yoma-mobile/blob/8a80c2ba5a4eedaec028e49e99daef02aadd8ca8/android/app/build.gradle#L143

cameronolivier (changes requested):

cameronolivier (changes requested):
You have a bunch of conflicts.

src/modules/App/App.middleware.ts
```diff
@@ -16,6 +17,7 @@ export const appResetFlow: Middleware =
     if (resetAppData.match(action)) {
       dispatch(AuthActions.clearAuth())
       dispatch(UserActions.clearUser())
+      dispatch(OrganisationsActions.clearOrganisations())
```
cameronolivier: suggestion: we shouldn't clear organisations as I don't believe they're tied to any user data.

src/modules/App/App.middleware.ts
```diff
@@ -16,6 +17,7 @@ export const appResetFlow: Middleware =
...
       dispatch(JobActions.clearJob())
```
cameronolivier: suggestion: Shouldn't clear jobs either

src/redux/middleware.ts
```diff
@@ -10,6 +10,7 @@ import { types as ApiUsersTypes } from '../api/users'
 import { middleware as AppMiddleware } from '../modules/App'
 import { middleware as AuthMiddleware } from '../modules/Auth'
 import { middleware as ErrorMiddleware } from '../modules/Error'
+import { middleware as JobsMiddleware } from '../modules/Jobs'
```
cameronolivier, line 13: Question: why are we adding jobs middleware in the `organisations` PR? Are you sure this is intended?

cameronolivier (changes requested):

src/redux/middleware.ts
sciphergfx: this is in develop @cameronolivier

cameronolivier: I looked at 'changes since last commit' - which I presume now was a merge/ conflict fix from develop. My bad.

src/modules/App/App.middleware.ts
sciphergfx: I'm lost here, this is  a different task if i'm not mistaken? @cameronolivier

src/modules/App/App.middleware.ts
sciphergfx: noted

src/modules/Organisations/Organisations.middleware.tsx
cameronolivier: question: @sciphergfx further to this - does this work? All the adjustments re normalising with 'key' prop are in the skills PR, so I don't think this will actually work yet?

cameronolivier (commented):
Left a question as I don't think the normalising will be functional until the skills work (And adjustments to switching between 'id' and 'key' are in.

src/modules/App/App.middleware.ts
cameronolivier: sorry @sciphergfx as mentioned - I filtered to see changes in last commit - but that was a merge in from develop -so it highlighted the jobs changes.

src/modules/Organisations/Organisations.middleware.tsx
sciphergfx: ok the skills PR is ready, we can merge that, then i adjust here @cameronolivier

sciphergfx: Let me fix conflicts there

cameronolivier (approved):

PR #106, opened 2021-08-06
YOMA-500 - refactor(skills): add skills data to state

sciphergfx:
## Work Done
- added skills data to state
- handles fetching skills only for user experience form


src/modules/Experience/Experience.selector.test.tsx
```diff
@@ -0,0 +1,18 @@
+import { rootStateFixture } from 'redux/redux.test.fixtures'
+
+import * as SUT from './Experience.selector'
+
+describe('modules/Experience/Experience.selector', () => {
+  describe('selector', () => {
```
cameronolivier: add another test please - for with data.

src/utils/strings.utils.ts
```diff
@@ -7,6 +8,17 @@ export const mapToDropDownArray = (array: Record<string, string>[], valueProp =
   }))
 }
 
+export const dropDownFromArray = (array: string[]) => {
```
cameronolivier: This should be moved (along with the mapToDropDownArray above) out of the strings utils. 

as should  `sliceArrayByNumber` and `searchArrayOfObjByValue`

None of these are utils that work on strings. 

Both `mapToDropDownArray` and `dropDownFromArray` - for one they seem to be doing the same thing (in the end) - but neither of them makes much sense to what they're doing by how they are named.

But they should be moved to the `DropDown` component utils - as they are utils for preparing the data for the DropDown component.

src/modules/Skills/Skills.middleware.tsx
```diff
@@ -0,0 +1,78 @@
...
+  fetchSkillsFailure,
+  fetchSkillsSuccess,
+  filterSkillsByName,
+  setFilteredSkills,
+  setSkillEntities,
+} from './Skills.reducer'
+import { selectSkillEntities } from './Skills.selector'
+import { extractSkillsFromPayload } from './Skills.utils'
+
+export const fetchSkillsFlow: Middleware =
```
cameronolivier, line 19: we should normalize the skills data - a key is the same as an ID - we might just need to slightly adjust the utils for normalizing so it pulls out the key instead of the ID.

src/modules/Skills/Skills.middleware.tsx
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+export const filterSkillsByNameFlow: Middleware =
+  ({ getState, dispatch }) =>
+  next =>
+  action => {
+    const result = next(action)
+    if (filterSkillsByName.match(action)) {
+      const state = getState()
+      const skillEntities = selectSkillEntities(state) as []
```
cameronolivier: we should know the shape of the skillsEntitites array, please type correctly.

src/utils/strings.utils.ts
```diff
@@ -7,6 +8,17 @@ export const mapToDropDownArray = (array: Record<string, string>[], valueProp =
...
+  return array.map((opt: string) => ({
+    label: opt,
+    value: opt,
+  }))
+}
+
+export const sliceArrayByNumber = (number: Number, array: any[]) => slice(0, number, array)
```
cameronolivier: slicing is generally by number, so the naming is a bit obtuse.

I'd suggest something that describes slicing the array to a point.

so maybe `sliceTo` or `sliceAt`?

src/utils/strings.utils.ts
```diff
@@ -7,6 +8,17 @@ export const mapToDropDownArray = (array: Record<string, string>[], valueProp =
...
+export const searchArrayOfObjByValue = (query: string, array: any[]) =>
```
cameronolivier: we are essentially filtering - so let's keep the same convention.

`filterByQuery` maybe?

src/modules/Skills/Skills.reducer.tsx
```diff
@@ -0,0 +1,32 @@
+import { createAction, createReducer } from '@reduxjs/toolkit'
+
+import { Skills, SkillsState } from './Skills.types'
+
+const name = '[Skills]'
+export const INITIAL_STATE = {
+  filteredSkills: [],
+  skillEntities: [],
```
cameronolivier: I'd expect skillEntities to follow a similar shape to the other `<credential>Entities` flow in:
```ts
{ someKey: { key: 'someKey', value: 'someValue' }
```

if we're filtering it might be useful to have an additional  entitites and all mapping as:
```ts
skillsByName: { skillValue: {key: 'skill key', value: 'skill value'}}
allSkillNames: ['skillValue1', 'skillValue2', ....]
```
presuming this would be simpler to filter. Then searching would simply search across an array of strings - which will be faster than an array of objects
and results would be mapping over the filters and pulling in the data by the value. 

....finally, I think it might be worthwhile finding a third-party auto-complete component? Not sure if rolling our own would be worth the hassle?

cameronolivier: Also this is the skills state key - so it feels redundant to repeat 'skills' in every prop name.
I think `filtered` and then `byValue` and `allValues` - and/or `byKey` and `allKeys`

To maintain the naming convention it might be useful to keep the naming convention here as `byId`- but that could cause confusion when looking at the data.

src/modules/Skills/Skills.middleware.test.tsx
```diff
@@ -0,0 +1,170 @@
...
+  setSkillEntities,
+} from './Skills.reducer'
+import { selectSkillEntities } from './Skills.selector'
+import { extractSkillsFromPayload } from './Skills.utils'
+
+describe('modules/Skills/Skills.middleware', () => {
+  describe('fetchSkillsFlow', () => {
+    it('should correctly handle being called', () => {
+      // given ...
+      const mockState = rootStateFixture({})
```
cameronolivier: no need to mock state for this one.

src/modules/Skills/Skills.middleware.test.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+      const create = createMiddlewareStub(jest, mockState)
+      const action = filterSkillsByName('QUERY')
+      // @ts-ignore
+      const { invoke, store } = create(SUT.filterSkillsByNameFlow)
+      // when ... we respond to the filterSkillsByName action
+      invoke(action)
+
+      // then ...validate filterSkillsByNameFlow
+      const state = store.getState()
```
cameronolivier: you're just redoing the code that here - so you're not really testing anything (you're just testing that the same/similar code written in 2 places does the same thing.

Rather mock out what you expect your code to look like when you run this middleware - and assert against that. 

So mock decent state data with a decent query value - and then assert that you're dispatching with the expected values.

src/modules/Skills/Skills.middleware.test.tsx
```diff
@@ -0,0 +1,170 @@
...
+        },
+      }
+
+      const action = fetchSkillsSuccess(mockResponseData)
+      // @ts-ignore
+      const { store, invoke } = create(SUT.fetchSkillsSuccessFlow)
+      // when ... we respond to the fetchSkillsSuccess action
+      invoke(action)
+      // then ...validate setSkillEntities
+      const skills = extractSkillsFromPayload(action)
```
cameronolivier: again - you're simply repeating code you've already created -hence testing nothing.

Please mock out the expected data in and the expected data out - and make sure that your middleware is doing what you presume it is.

cameronolivier (changes requested):

src/modules/Skills/Skills.reducer.tsx
sciphergfx: there's a solution in place that handles this autocomplete part on the frontend. It's tied to the reducer, have you had a look?

sciphergfx: I think this may also apply to organisations in the long-run, we might need to standard approach on for that.

Do you reckon we stick to this as the state shape

skillsByName: { skillValue: {key: 'skill key', value: 'skill value'}}
allSkillNames: ['skillValue1', 'skillValue2', ....] // this should handle filtered?
skillEntities: { someKey: { key: 'someKey', value: 'someValue' }}

Please let me know what you think

src/modules/Skills/Skills.middleware.test.tsx
sciphergfx: I'll attend to this, just need an idea on that state shape so I can replicate where needed

src/utils/strings.utils.ts
```diff
@@ -1,18 +1,19 @@
-import { head, join, juxt, pipe, tail, toUpper, trim, filter } from 'ramda'
+import { filter, head, join, juxt, pipe, slice, tail, toUpper, trim } from 'ramda'
 
 export const mapToDropDownArray = (array: Record<string, string>[], valueProp = 'key', labelProp = 'value') => {
   return array.map((opt: Record<string, string>) => ({
     label: opt[labelProp],
     value: opt[valueProp],
   }))
 }
+export const sliceAt = (number: Number, array: string[]) => slice(0, number, array)
```
cameronolivier: this should not be in the strings utils. probably an array utils - or the ramda utils.

src/utils/strings.utils.ts
```diff
@@ -1,18 +1,19 @@
...
 
 export const getUppercasedHead = (data: string) => pipe(trim, head, toUpper)(data)
 
 export const capitalize = pipe(juxt([pipe(head, toUpper), tail]), join(''))
 
 export const filterStringArray = (value: string, array: string[]) =>
```
cameronolivier: also should be in the array utils - not in strings.

src/modules/Skills/Skills.types.tsx
```diff
@@ -0,0 +1,18 @@
...
+}
+
+export type SkillsState = {
+  filtered: string[]
+  allValues: string[]
+  allKeys: string[]
+  byValue: Record<string, Skills>
+  byKey: Record<string, Skills>
+}
+export type NormalisedSkillsStatse = {
```
cameronolivier: spelling mistake here

src/modules/App/App.middleware.ts
```diff
@@ -28,6 +30,7 @@ export const hydrateAppFlow: Middleware =
     const result = next(action)
     if (hydrateApp.match(action)) {
       dispatch(UserActions.fetchUserCredentials())
+      dispatch(SkillsActions.fetchSkills())
```
cameronolivier, line 32: can you update the tests for both of these.

src/modules/Experience/Experience.selector.test.tsx
cameronolivier: Why did you remove the one without data? 
We need both initial state and with data tests for selectors.

src/modules/Experience/Experience.selector.test.tsx
```diff
@@ -0,0 +1,26 @@
...
+          ...SKILLS_STATE_MOCK,
+          filtered: mockFiltered,
+        },
+      })
+
+      // when ... we call the selector
+      const result = SUT.default(state)
+
+      // then ... should return filtered skills
+      const { skills } = result
```
cameronolivier: please do this the other way around. We want to test that result equals {skills: mockFiltered}. not that the skills prop is correct-  but that the entire selector is as expected.

src/modules/Skills/Skills.reducer.tsx
```diff
@@ -0,0 +1,37 @@
...
+  allValues: [],
+  allKeys: [],
+  byValue: {},
+  byKey: {},
+} as SkillsState
+
+export const setSkillEntities = createAction<NormalisedSkillsState>(`${name} setSkillEntities`)
+export const setFilteredSkills = createAction<string[]>(`${name} setFilteredSkills`)
+
+export const filterSkillsByName = createAction<string>(`${name} filterSkillsByName`)
```
cameronolivier: I don't think this is used anywhere? 
Or let me rephrase - you have middleware that catches it -but nowhere that's dispatching it.

src/modules/Skills/Skills.middleware.tsx
```diff
@@ -0,0 +1,80 @@
...
+            onSuccess: fetchSkillsSuccess,
+            onFailure: fetchSkillsFailure,
+          }),
+        ),
+      )
+    }
+    return result
+  }
+
+export const filterSkillsByNameFlow: Middleware =
```
cameronolivier: as the action is never dispatched this will never run.

cameronolivier: please remove this middleware - and the corresponding action

src/modules/Skills/Skills.middleware.tsx
```diff
@@ -0,0 +1,80 @@
...
+    return result
+  }
+
+export const fetchSkillsSuccessFlow: Middleware =
+  ({ dispatch }) =>
+  next =>
+  action => {
+    const result = next(action)
+
+    if (fetchSkillsSuccess.match(action)) {
```
cameronolivier: please see the UserChallenge module for an example on handling getting data, normalising in middleware and setting to state.

src/modules/Skills/Skills.reducer.tsx
```diff
@@ -0,0 +1,37 @@
...
+export const fetchSkillsFailure = createAction<string>(`${name} fetchSkillsFailure`)
+export const clearSkills = createAction(`${name} clearSkills`)
+
+const SkillsReducer = createReducer(INITIAL_STATE, builder => {
+  builder.addCase(setSkillEntities, (state, action) => ({
+    ...state,
+    ...normaliseByKey(action.payload),
+    ...normaliseByValue(action.payload),
+  }))
+  builder.addCase(setFilteredSkills, (state, action) => ({
```
cameronolivier: should be handled by the selector

cameronolivier (changes requested):
Some redux rules of thumb:
don’t store data in state that can be derived from the existing data.
use selectors to manipulate and derive the data you need for the component.

Then these are for use in our project:
don’t do data manipulation in reducers - do it in the middleware
stick to our convention for normalised state: {ids: string[], entities: {}} . you can swap ids out with keys - or another unique identifier depending on the data.

Don’t store data in state that can be derived from the existing data.
Store the data in skills state once. Have a selector that manipulates it if need be to what you require for usage in the component. In this case also don’t store filtered data in state - as we’re just slicing data. build this into a selector. 

Use selectors to manipulate and derive the data you need for the component.
Following on from above - use a selector to take the state data and derive it into how you might need it in the component - if that’s pulling out all values, or slicing for top 20, or whatever else you might need.

Don’t do data manipulation in reducers - do it in the middleware
Following the above 2 would solve this problem - but the way we’ve architected the app at the moment - things like normalising the data occurs in the middleware, not in the reducer.

src/modules/Experience/Experience.selector.test.tsx
sciphergfx: As per comments from other PRs, I'm trying to stick with what the task says "add skills data to state", I'll add selectors when we get to relevant stories

src/utils/redux.utils.ts
```diff
@@ -1,9 +1,9 @@
-import { applySpec, identity, keys, path, pipe, prop } from 'ramda'
+import { applySpec, has, identity, ifElse, keys, path, pipe, prop } from 'ramda'
 
 import { objFromListWith } from './ramda.utils'
 
 export const normalise = pipe(
-  objFromListWith(prop('id')),
+  objFromListWith(ifElse(has('key'), prop('key'), prop('id'))),
```
cameronolivier: ### Issue: should avoid longer, nested functions

### Suggestion:
I think it's a simpler, and more straightforward way to simply pass the prop, with a default value:
```ts
export const normalise = (data, identifier = 'id') => pipe(
  objFromListWith(prop(identifier)),
```
and use it like so:
```ts
normalise(data)
// OR:
normalise(data, 'key')
```

cameronolivier (changes requested):
small change. otherwise looking good! :)

src/utils/redux.utils.ts
```diff
@@ -2,12 +2,13 @@ import { applySpec, identity, keys, path, pipe, prop } from 'ramda'
...
 import { objFromListWith } from './ramda.utils'
 
-export const normalise = pipe(
-  objFromListWith(prop('id')),
-  applySpec({
-    entities: identity,
-    ids: keys,
-  }),
-)
+export const normalise = (data: any, identifier = 'id') =>
```
cameronolivier: nitpick: data type should be:
```ts
data: Record<string, any>[]
```

src/types/general.types.ts
```diff
@@ -1,2 +1,2 @@
 export type StdObj<T = any> = Record<string, T>
-export type StdFn<T = any, R = void> = (args: T) => R
+export type StdFn<T = any, R = void> = (args: T, identifier?: string) => R
```
cameronolivier, line 2: issue: we shouldn't adjust this - as it's meant to be a generic function type, so setting the second argument means it can't be used elsewhere.

The original is typed so that it can only take 1 argument, which limits its utility.

suggestion: 
I can think of 3 decent options - feel free to use or not use:
**Option 1:** Manually type it in the `skills.middleware`: 
```ts
(payload: Skill[], identifier: string) => NormalisedSkills
```

**Option 2:** Create a new generic type that takes 2 arguments:
```ts
export type StdFn2Args<Arg1 = any, Arg2 = any, Return = void> = (arg1: Arg1, arg2: Arg2) => Return
```
usage in the Skills middleware:
```ts
StdFn2Args<Skill[], string, NormalisedSkills>
```
This will work, but obviously only works with functions with 2 arguments. 

**Option 3:** create a generic type that takes `n` arguments:
```ts
export type StdFnWithArgs<Args = any, Return = void> = (...args: Args[]) => Return
```
usage in the Skills middleware:
```ts
StdFnWithArgs<Skill[] | string, NormalisedSkills>
```
_Issue is we aren't STRICTLY typing the inputs here, as in you could switch the action.payload and the key around in the call and typescript wouldn't throw an error, so options 1  or 2 would be more precise_ 
ie this wouldn't throw a typescript error, but it would cause the function to error:
```ts
 normalise('key', action.payload)
```

cameronolivier (changes requested):

src/types/general.types.ts
sciphergfx: So since the function is instantiated with  arg "id", that will always be a constant hence identifier being optional. Tried option 2 before taking that position. Happy to refactor otherwise, Which option do you advice I effect just so we're are on the same page? @cameronolivier

src/modules/App/App.middleware.ts
```diff
@@ -16,6 +17,7 @@ export const appResetFlow: Middleware =
     if (resetAppData.match(action)) {
       dispatch(AuthActions.clearAuth())
       dispatch(UserActions.clearUser())
+      dispatch(SkillsActions.clearSkills())
```
cameronolivier: issue: Skills should not be cleared (arguably ever).

cameronolivier (changes requested):

src/types/general.types.ts
cameronolivier: @sciphergfx 

The primary thing is we shouldn't be changing a generic type for a specific use-case.

On typing the normalise function in skills middleware: In that instance it does take 2 args, even though the second is optional.

A 4th option might be to create an actual type for the `normalise` function, that you then simply use everywhere you want to import `normalise`. 

```ts
export type Normalise<Data, NormalisedData> = (data: Data[], identifier?: string) => NormalisedData
```
usage in the Skills middleware:
```ts
Normalise<Skill, NormalisedSkills>
```

If I were doing it, I'd probably go with the above option. Otherwise, I'd go with 2, or 1 (2 is slightly better than 1 imo, because it has more utility down the line)

src/types/general.types.ts
```diff
@@ -1,2 +1,3 @@
...
+export type Normalise<Data, NormalisedData> = (data: Data[], identifier?: string) => NormalisedData
```
cameronolivier, line 3: this is great - but I think it should be in a different file. Happy for you to do this in the next PR so we can get this in. 

I'd suggest creating a `redux.types.ts` file in this folder and adding it there.

cameronolivier (approved):
Just a comment to move that new Normalise type -but can do in Organisations PR.

src/types/general.types.ts
sciphergfx: We can merge, waiting to effect some changes on organisations PR @cameronolivier

PR #107, opened 2021-08-07
YOMA-451 - feat(Experience): experience should correctly show all the user s entries

sciphergfx:
## Work Done
- Show all user jobs entries
## Screenshot
Simulator Screen Shot - iPhone 8 - 2021-08-22 at 17 32 27


src/modules/Experience/Experience.selector.tsx
```diff
@@ -0,0 +1,6 @@
+import { selectJobEntities } from 'modules/Job/Job.selector'
```
cameronolivier: should we not be getting the jobs from the user credentials - not from the Job modules-  which would be all the jobs.

src/modules/Experience/Experience.container.tsx
```diff
@@ -0,0 +1,20 @@
+import { Dispatch } from '@reduxjs/toolkit'
+import { actions as JobActions, types as JobTypes } from 'modules/Job'
+import { connect } from 'react-redux'
+
+import Experience from './Experience'
+import selector from './Experience.selector'
+
+const mapStateToProps = selector
+const mapDispatchToProps = (dispatch: Dispatch) => {
```
cameronolivier: this story just cares about _displaying_ the experience data - not about any of the editing. You've added a bunch of code aimed at handling the editing part - which should be extracted into `452` - where it's meant to be handled.

src/modules/Experience/Experience.selector.tsx
cameronolivier: if this is the credentials, then it should be inside the User module as discussed.

src/modules/Experience/Experience.tsx
```diff
@@ -1,65 +1,116 @@
...
 import styles from './Experience.styles'
-import { ExperienceType } from './Experience.types'
+import { ExperienceFormState, ExperienceType } from './Experience.types'
+import { extractExperienceFormValues } from './Experience.utils'
 import ExperienceForm from './ExperienceForm/ExperienceForm'
+import { INITIAL_VALUES } from './ExperienceForm/ExperienceForm.constants'
+import { DropDownList } from './ExperienceForm/ExperienceForm.types'
 
 interface Props {
+  onJobCreate: (job: JobRequestPayload) => void
```
cameronolivier: as mentioned above we should only be handling data required for displaying the user's experiences in this story. I presume that would simply be jobs.

src/modules/Experience/Experience.tsx
```diff
@@ -1,65 +1,116 @@
...
+  onJobUpdate: (job: JobUpdatePayload) => void
+  filterSkillsByName: (query: string) => void
+  jobs: []
```
cameronolivier: please type this correctly.

src/modules/Experience/Experience.tsx
```diff
@@ -1,65 +1,116 @@
...
   const { t } = useTranslation()
   const [isSaved, setIsSaved] = useState(false)
-  const [experience] = useState([])
-  const formRef = useRef<FormikProps<FormikValues>>()
+  const [isEditMode, setIsEditMode] = useState(false)
+  const [jobsList, setJobsList] = useState([])
+  const [formState, setFormState] = useState<ExperienceFormState>({ isValid: true, values: INITIAL_VALUES })
+
+  useEffect(() => {
+    setJobsList(Object.values(jobs))
```
cameronolivier: please set the data you need in the selector - that's where we should be preparing the data for a component - not in the component itself.

Also - we should have these normalised.

src/modules/Experience/Experience.tsx
```diff
@@ -1,65 +1,116 @@
...
             fallback={<EmptyCard title={t('Where do you currently work?')} onAdd={() => setIsSaved(true)} />}
           >
             <FlatList
-              data={experience}
+              data={jobsList}
               contentContainerStyle={styles.listContainer}
-              renderItem={({ item }) => renderItem(item)}
-              keyExtractor={(item, index) => index.toString()}
+              renderItem={({ item }: any) => renderItem(item)}
+              keyExtractor={(item: any, index: number) => index.toString()}
```
cameronolivier: this shouldn't be necessary, I don't think - but also - don't all credentials have ids? so our list of experience should have an id attached - if it's not pulling it in automatically, then we should be getting it from there.

cameronolivier (changes requested):
This feature is meant to handle displaying data - so should arguably have the middleware that pulls the data out of state and saves into middleware. 

It should not have anything that handles adding/saving new experience. 

I'd suggest refactoring between this and 452 so that this lays the foundation for getting the data into state and 452 handles saving editing the data.

It will make this a bit bigger (which is perfectly fine - it's only 8 files) and will reduce 452 in size which is great as it's still 50 files large.

Also if we do it that way, this should be able to go in first and separate to 452. 

And 452 can then only be concerned with saving or editing a specific user experience instead of trying to handle all of these things in one massive chunk of work.

src/modules/UserJobs/UserJobs.types.ts
```diff
@@ -39,6 +39,11 @@ export interface NormalisedUserJobs {
   entities: Record<string, UserJobCredential>
 }
 
+export type UserJobsList = {
```
cameronolivier: suggestion: I think this should maybe be `UserJobItem`? it's not a list - it's an item (that may or may not be in a list) 

IF we need to differentiate between this and the widget or other UserJobItems, then maybe UserJobScreenItem or UserJobWidgetItem for the widget data.

cameronolivier (changes requested):

src/modules/UserJobs/UserJobsForm/UserJobsForm.constants.ts
```diff
@@ -10,6 +10,28 @@ export const INITIAL_VALUES: UserJobFormFields = {
   endTime: null,
   published: false,
 }
+export const MOCKED_JOBS = [
```
cameronolivier: issue: you can remove this - it's not used at all in the project.

src/modules/UserJobs/UserJobs.tsx
```diff
@@ -46,7 +51,7 @@ const UserJobs = ({ onJobCreate, onJobPatch, filterSkillsByValue, navigation }:
   }
 
   const renderItem = (item: any) => {
```
cameronolivier, line 48: suggestion: I think we should move this outside the component. Happy to keep it in the same file (although it might be best to extract into its own component).

src/modules/UserJobs/UserJobs.tsx
```diff
@@ -14,19 +14,24 @@ import UserJobsForm from './UserJobsForm/UserJobsForm'
...
+  navigation: StackNavigationProp<HomeNavigatorParamsList, HomeNavigationRoutes.UserJobs>
 }
 
-const UserJobs = ({ onJobCreate, onJobPatch, filterSkillsByValue, navigation }: Props) => {
+const UserJobs = ({ userJobs, onJobCreate, onJobPatch, filterSkillsByValue, navigation }: Props) => {
   const { t } = useTranslation()
   const [isSaved, setIsSaved] = useState(false)
   const [isEditMode, setIsEditMode] = useState(false)
-  const [userJobsList] = useState<[]>([])
+  const [userJobsItems, setUserJobItems] = useState<UserJobsTypes.UserJobItem[]>([])
```
cameronolivier: issue: this serves no purpose.
suggestion - remove this and simply inject `userJobs` directly into the flatlist data.

src/modules/UserJobs/UserJobs.selector.tsx
```diff
@@ -1,6 +1,15 @@
 import { createSelector } from '@reduxjs/toolkit'
-import { prop } from 'ramda'
+import { map, pick, pipe, prop, values } from 'ramda'
 import { RootState } from 'redux/redux.types'
 
 export const selectUserJobs = (state: RootState) => state.userJobs
 export const selectFormValues = createSelector(selectUserJobs, prop(['formValues']))
+
+export const selectUserJobItem = createSelector(
```
cameronolivier: nitpick - I was going to say that this should be plural (`selectUserJobItems`) - but maybe it's best to simply make this the default selector and then no need to name it at all. 

Essentially you're wanting to select the user jobs formatted for the user job item. but that's a little verbose - and as we're using the default selector I don't see any reason not to simply export this all as a single default selector. 

just add `objOf({userJobs})` to the end of the pipe on line 10.

cameronolivier (changes requested):

src/modules/UserJobs/UserJobs.selector.tsx
cameronolivier: suggestion: Forgot to add -please remember to bump version and release notes.

sciphergfx: A bit puzzled with regards the use of objOf({userJobs}), props needed on the UI is {'job', 'startDate', 'endDate'}?

src/modules/UserJobs/UserJobs.tsx
sciphergfx: Our component has less than 150 lines, and I doubt it would be re-used, shall I extract nonetheless?

cameronolivier: you can keep it there for now.

cameronolivier (approved):

PR #108, opened 2021-08-08
YOMA-497 - feat(UserChallenges): add data to state

cameronolivier:
## Work Done
- adding flows to:
  - get challenges from credentials response
  - normalise the data
  - set the data to state 
- setup reducer - adding actions and reducer to handle data flow and store data

### STILL TODO:
- finish sorting out tests
- type challenges data correctly 
- finish adding middleware from UserCredentials to middleware

## Steps to Test
- Run tests


src/modules/UserChallenges/UserChallenges.middleware.ts
```diff
@@ -0,0 +1,48 @@
...
+
+import { StdFn } from '../../types/general.types'
+import * as UserActions from '../User/User.reducer'
+import { UserCredentials } from '../User/User.types'
+import { getUserChallengesSuccess, normaliseUserChallengesSuccess, setUserChallenges } from './UserChallenges.reducer'
+import { NormalisedUserChallenges, UserChallenge } from './UserChallenges.types'
+
+export const getUserChallengesFromCredentialsFlow =
+  (
+    extractDataFromPayload: StdFn<any, UserCredentials>,
```
sciphergfx, line 11: I think I created a "extractDataFromPayload" util in YOMA-452, but we can always merge

src/modules/UserChallenges/UserChallenges.middleware.ts
```diff
@@ -0,0 +1,48 @@
...
+    extractChallenges: StdFn<UserCredentials, UserChallenge[]>,
```
sciphergfx, line 12: If we're going to introduce a new approach of adding util functions as dependencies via middleware, then we need to evaluate the time cost of standardising this app wide. Otherwise, happy with it.

src/redux/middleware.ts
```diff
@@ -48,6 +51,12 @@ const featureModuleMiddleware = [
   AuthMiddleware.setSecureRefreshTokenFlow(SecureStore.setItemAsync),
   AuthMiddleware.unauthorizedFlow,
   ErrorMiddleware.categorizeErrorsFlow,
+  UserChallengesMiddleware.getUserChallengesFromCredentialsFlow(
```
sciphergfx, line 54: commented on this approach above

sciphergfx, line 54: commented on this approach above

src/modules/UserChallenges/UserChallenges.types.ts
```diff
@@ -0,0 +1,37 @@
...
+  createdByAdmin: boolean
+  language: string
+  startTime: string
+  endTime: string | null
+  published: boolean
+}
+
+export type UserChallenge = {
+  challenge: Challenge
+  id: string
```
sciphergfx: So all props below the challenge prop represents user credentials meta, shared by other credentials type like jobs, etc. Why not separate the type to avoid repetition?

I did 


```js
export interface UserCredentialMeta {
  id: string
  verifiedAt: string
  approved: boolean
  approvalMessage: string
  startDate: string
  endDate: string
  createdAt: string
  fileId: string
  fileURL: string
  requestVerification: boolean
}
```

then extended the interface. Your thoughts?

src/modules/UserChallenges/UserChallenges.middleware.ts
cameronolivier: you did :) but I needed something now - so figured it would be easier if I started using a shared util now than remake my own. And can merge :)

src/modules/UserChallenges/UserChallenges.middleware.ts
cameronolivier: I did this because it makes it easier to specifically test what we need the middleware to do. We are already testing the utils themselves, so it makes sense to isolate them and only include if we specifically need to. 

I don't think this needs a site-wide refactor, but we can refactor as we find it appropriate. All current implementations are already tested - so I don't think it's necessary to set aside time for such a refactor tbh.

src/modules/UserChallenges/UserChallenges.types.ts
cameronolivier: yeah I'm happy with that - I wasn't sure that the data shape persisted across credentials (although in hindsight that makes sense that it would - that's a good approach, thanks! I presume this is in User.types?

sciphergfx: yes it is

sciphergfx (approved):
All good

PR #109, opened 2021-08-09
YOMA-501 - feat(UserJobs): added jobs data to state

sciphergfx:
## Work Done
- added jobs data to state
- refined normalise util function to accept object


src/api/api.utils.ts
```diff
@@ -30,10 +32,15 @@ export const prependIdToEndpointInConfig = (config: Partial<ApiMeta>) => (id: st
   evolve({
     endpoint: pipe(of, concat([id])),
   })(config)
+export const appendIdToEndpointInConfig = (config: Partial<ApiMeta>) => (id: string) =>
```
cameronolivier: needs tests added

src/api/api.utils.ts
```diff
@@ -30,10 +32,15 @@ export const prependIdToEndpointInConfig = (config: Partial<ApiMeta>) => (id: st
...
+  evolve({
+    endpoint: append(id),
+  })(config)
 export const addParamsToConfig = addValueWithGivenKeyToConfig('params')
 export const createParam = objOf
 export const createTypeParam = createParam('type')
 
+export const extractPayloadData = path(['payload', 'data', 'data'])
```
cameronolivier: need tests added

src/modules/Error/Error.utils.ts
```diff
@@ -0,0 +1,3 @@
+import { path } from 'ramda'
+
+export const extractErrorMessageFromPayload = path(['payload', 'data', 'meta', 'message'])
```
cameronolivier: I'd think we might need 2 utils - 1 to extract the meta (although we might have that already?) and then a second to access the message in meta? or what do you think?

src/modules/Experience/Experience.types.ts
```diff
@@ -1,15 +1,14 @@
...
 
-type Job = {
-  description: string
-  id: string
-  organisationId: string
-  organisationLogoURL: string
-  organisationName: string
-  title: string
+export type ExperienceFormState = {
+  values: FormikValues
```
cameronolivier: FormikValues seems to be very generic. surely we need more typing here? I'm presuming values for here would be different to values for another form?

src/modules/Job/Job.middleware.test.tsx
```diff
@@ -0,0 +1,501 @@
...
+import { selectJobCredentialIdFromTmpFormValues, selectJobTmpFormValues } from './Job.selector'
+import { defaultJobsResponseData } from './Job.test.fixtures'
+import { JobCredentialsTmpFormValues } from './Job.types'
+import {
+  extractJobCredentialId,
+  extractJobCredentialRequestPayload,
+  extractJobCredentialUpdatePayload,
+} from './Job.utils'
+
+describe('modules/Jobs/Jobs.middleware', () => {
```
cameronolivier: should be singular

src/modules/Job/Job.middleware.tsx
```diff
@@ -0,0 +1,291 @@
+import { extractPayloadData } from 'api/api.utils'
```
cameronolivier: so it seems you're doing both global jobs and specific user experience calls within the one module - I think it should be separated out, as discussed, into 
- `Jobs` - which handles all jobs (via `​/api​/v1​/jobs...` api calls) and 
- `User/Jobs` or the more I think of it `User/Experience` which handles all of the User's experience/jobs and is managed by `​/api​/v1​/users​/{userId}​/credentials...` api calls.

cameronolivier: This comment applies for this entire Jobs module that you created as well as any other place where credentials assigned to a user are managed separately to all possible entities for a given category (Skills, Challenges, etc)

cameronolivier (changes requested):
Job or Jobs should be inside `User` module if it's a user credential, as discussed.

If it's just a list of data (jobs/organisations, etc) _NOT_ tied to the user then it should have a root-level module.

src/api/api.utils.ts
sciphergfx: This is in YOMA-452, remember it was extracted, but i'll add

src/api/api.utils.ts
sciphergfx: same as above

src/modules/Error/Error.utils.ts
sciphergfx: noted

src/modules/Experience/Experience.types.ts
sciphergfx: I added this to YOMA-451, I realised this is the JobRequestPayload type present there

src/modules/Job/Job.middleware.tsx
sciphergfx: I understand you. So skills is not a user credential per-say, but compliments them. The API handles it differently if I'm not mistaken?

sciphergfx:
re-open

src/api/api.utils.test.ts
```diff
@@ -82,6 +101,27 @@ describe('api/api.utils', () => {
       expect(result).toEqual({ someKey: 'someValue' })
     })
   })
+
+  describe('extractPayloadData', () => {
```
cameronolivier: this should now be in develop :) 

I added it to the redux utils because we're extracting from an action.... but maybe it should actually be in api utils because it's not general extraction - it's typical to api response data....

So if you feel like moving that - go for it - but otherwise use it from `src/utils/redux.utils.ts`

src/api/api.utils.ts
cameronolivier: as mentioned in the tests -this is now in the project in `src/utils/redux.utils.ts` - I think that should be moved. Either here - or to `src/utils/api.utils.ts`

src/modules/App/App.middleware.ts
```diff
@@ -1,5 +1,6 @@
 import { Middleware } from 'redux'
 
+import * as JobsActions from '../UserJobs/UserJobs.reducer'
```
cameronolivier: this should be `UserJobsActions`.

src/modules/Experience/Experience.types.ts
```diff
@@ -1,15 +1,8 @@
+import { UserJobsOrganisationPayload } from 'modules/UserJobs/UserJobs.types'
+
 export type ExperienceType = {
   endDate: string
   id: string
-  job: Job
+  job: UserJobsOrganisationPayload
```
cameronolivier: If this is just an Organisation, then let's call it that. If it's a JobsOrganisation, then let's call it that - I think ONLY add 'Payload' if it's specifically a payload shape - and we're using a different shape in the app. When it's actual data that we're using around the app, adding 'Payload' is a bit redundant -and unnecessarily verbose.

So - as I have a `Challenge` - which is the `challenge: {...}` part of the data and UserChallenge which is the `Challenge` + `UserCredentialMeta` - could you not follow a similar naming pattern? 

And further why would the Experience > job be a UserJobsOrgansationPayload? The naming doesn't seem to map. Surely it should be something like `job: UserJob`?

src/modules/User/User.middleware.ts
```diff
@@ -201,3 +206,31 @@ export const updateUserPhotoFailureFlow =
...
+
+    if (fetchUserCredentialsSuccess.match(action)) {
+      const userCredentialPayload = extractPayloadData(action)
+      const jobs = extractCredentialsByType(UserCredentialTypes.Job)(userCredentialPayload)
+      dispatch(setUserJobsEntities(normalise(jobs)))
+    }
+    return result
+  }
+
+export const fetchUserCredentialsFailureFlow =
```
cameronolivier: We do need this - thank you! that's a good catch :) 

Could we move this next to the fetchUserCredentialsFlow middleware (Around line 80) - so that they're grouped-  otherwise it's hard to see if we're missing something or the flow of things.

src/modules/User/User.middleware.ts
```diff
@@ -201,3 +206,31 @@ export const updateUserPhotoFailureFlow =
     }
     return result
   }
+
+export const fetchUserCredentialsSuccessFlow: Middleware =
```
cameronolivier: please follow my flow - and have middleware to extract the jobs in your UserJobs module. 

The User module shouldn't care about what other module needs what data in a success payload it's firing off. 

Right now the User module doesn't actually need to do any handling of the data (except if you simply want to extract the payload data and pass that on - that would be useful - but then it should dispatch a new action, something like `userCredentials` with the extracted userCredentials from the payload.

So the body of this middleware would simply be something like:
```ts
if (fetchUserCredentialsSuccess.match(action)) {
    const userCredentialData = extractPayloadData(action)
    dispatch(userCredentials(userCredentialData))
  }
```

And then we'd need to update my UserChallenges module to listen for that action instead.

src/modules/User/User.middleware.ts
```diff
@@ -201,3 +206,31 @@ export const updateUserPhotoFailureFlow =
...
+  ({ dispatch }) =>
+  next =>
+  action => {
+    const result = next(action)
+
+    if (fetchUserCredentialsSuccess.match(action)) {
+      const userCredentialPayload = extractPayloadData(action)
+      const jobs = extractCredentialsByType(UserCredentialTypes.Job)(userCredentialPayload)
```
cameronolivier: so this needs to move to the UserJobs middleware.

See my Challenges example.

src/modules/User/User.middleware.ts
```diff
@@ -201,3 +206,31 @@ export const updateUserPhotoFailureFlow =
...
+      dispatch(setUserJobsEntities(normalise(jobs)))
```
cameronolivier: Please don't call a function on the data inside an action dispatch. Call it separately - and preferably in its own middleware.

src/modules/UserJobs/UserJobs.middleware.tsx
```diff
@@ -0,0 +1,291 @@
...
+import { Middleware } from 'redux'
+import { showSimpleMessage } from 'utils/error'
+import { normalise } from 'utils/redux.utils'
+
+import { actions as ApiActions, utils as ApiUtils } from '../../api'
+import { constants as ApiUserJobsConstants } from '../../api/jobs'
+import { constants as ApiUsersConstants } from '../../api/users'
+import * as Navigation from '../Navigation/Navigation.actions'
+import {
+  createUserJobs,
```
cameronolivier: I think this middleware is doing too much for a 'add [User]Jobs data to state' task. 

See my Challenges - it's a good overview of what we need to be doing in the credentials. 

If it grows as we add editing & saving (in 452) then obviously that's great - but it shouldn't be in this PR

cameronolivier: None of the current middleware is required for this story. And the middleware that should be here isn't. Please refactor.

src/modules/UserJobs/UserJobs.reducer.tsx
```diff
@@ -0,0 +1,52 @@
+import { createAction, createReducer } from '@reduxjs/toolkit'
```
cameronolivier: there's also too much going on here for simply adding UserJobs data to state. 

We simply need actions that handle whatever you need to do that. Adn state should reflect that too - I'm not sure what tmpFormValues are - but they won't be needed here. 

I think all you need right now is 
```ts
{
 ids: [],
 entities: {},
}
```

cameronolivier: but - do what you need specifically to Add USER Jobs data to state.

If it's _all_ jobs - that needs to move to the Jobs module. 
If it's anything to do with managing the form or adding or editing user experience-  that needs to be in 452 - or other stories.

All this story/task should care about is pulling the UserJobs data out of the `fetchUserCredentialsSuccess` action payload, normalising it correctly and saving it to state. Nothing more, nothing less.

src/modules/UserJobs/UserJobs.selector.tsx
```diff
@@ -0,0 +1,11 @@
+import { createSelector } from '@reduxjs/toolkit'
```
cameronolivier: we don't need a selector yet.

src/modules/UserJobs/UserJobs.test.fixtures.ts
```diff
@@ -0,0 +1,56 @@
...
+    organisationURL: 'URL',
+    organisationPrimaryContactName: 'CONTACT_NAME',
+    organisationPrimaryContactEmail: 'EMAIL',
+    organisationPrimaryContactPhone: 'PHONE',
+    skills: ['SKILL'],
+    countries: ['COUNTRY'],
+  },
+}
+
+export const jobCredentialsStateData = {
```
cameronolivier: this shouldn't be a fixture - it has little bearing on real data - it's really just a mock - which can be inline (And should be inline) - Tests are one place where you don't need to be DRY - it's better to keep data that you're testing as close to the test code as possible - so only abstract if it's really bulky and/or fixture data (as in mocked actual response data) - (as above)

src/modules/UserJobs/UserJobs.types.tsx
```diff
@@ -0,0 +1,59 @@
+import { ApiMetaResponse } from 'modules/Auth/Auth.types'
+import { UserCredentialMeta } from 'modules/User/User.types'
```
cameronolivier: this doesn't exist on `User/User.types` according to my IDE (I double-checked and I can't find it either)

src/modules/UserJobs/UserJobs.types.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+export interface UserJobsPayload {
```
cameronolivier: remove 'Payload' if it's being used as a UserJob in the app - Payload adds nothing here (and can cause confusion)

src/modules/UserJobs/UserJobs.types.tsx
```diff
@@ -0,0 +1,59 @@
...
+  title: string
+  description: string
+  createdAt: string
+  createdByAdmin: boolean
+  language: string
+  published: boolean
+  skills: string[]
+}
+
+export interface UserJobsOrganisationPayload extends UserJobsPayload {
```
cameronolivier: I think this can/should be named better.

src/modules/UserJobs/UserJobs.types.tsx
```diff
@@ -0,0 +1,59 @@
...
+  organisationId: string
+  organisationName: string
+  organisationLogoURL: string
+  organisationURL: string
+  organisationPrimaryContactName: string
+  organisationPrimaryContactEmail: string
+  organisationPrimaryContactPhone: string
+}
+
+export interface UserJobsCredentialsTmpFormValues {
```
cameronolivier: what is this? also I don't think we need this yet.

src/modules/UserJobs/UserJobs.types.tsx
```diff
@@ -0,0 +1,59 @@
...
+  credentialId?: string
+  requestVerification?: boolean
+  startTime: string
+  endTime: string
+}
+
+export interface UserJobsResponsePayload {
```
cameronolivier: so this seems to be correctly named as a Payload.

cameronolivier: That said, we could probably remove 'Payload' as it actually looks like a Response. We don't need to postfix Payload to everything (unless it's ONLY a Payload data shape AND removing the Payload would clash with a different type)

Basically - never prefix Payload unless it makes a LOT of sense and there's no other option.

src/modules/UserJobs/UserJobs.types.tsx
```diff
@@ -0,0 +1,59 @@
...
+  endTime: string | null
+}
+
+export interface UserJobsUpdatePayload extends UserJobsRequestPayload {
+  id: string
+  credentialsId: string
+}
+
+export interface UserJobsCredentials extends UserCredentialMeta {
+  job: UserJobsOrganisationPayload
```
cameronolivier: this should really just be `Job` - unless there's another `Job` interface that is considerably different.

src/modules/UserJobs/UserJobs.types.tsx
```diff
@@ -0,0 +1,59 @@
...
+}
+export interface UserJobsState {
+  tmpFormValues?: any
```
cameronolivier: please adjust this to reflect what the state should be on fetching Jobs data - which should just be normalised jobs 
```ts 
{
id: string[],
entities: Record<string, UserJobs[]> 
}
```
presuming `UserJobs` is the type for the UserJob (I think you called it `UserJobsCredentials`, but I think `UserJob` is probably sufficient, and consistent with the Challenge naming convention. At any rate keep the naming convention singular if it's typing a single entity. (so `UserJob` not `UserJobs`).

src/utils/redux.utils.ts
```diff
@@ -1,8 +1,9 @@
-import { applySpec, identity, keys, pipe, prop } from 'ramda'
+import { applySpec, identity, ifElse, is, keys, of, pipe, prop } from 'ramda'
 
 import { objFromListWith } from './ramda.utils'
 
 export const normalise = pipe(
+  ifElse(is(Array), identity, of),
```
cameronolivier: we should only be normalising arrays of objects - that's what normalising does - so there's no need to be catering for non-arrays here.

src/redux/middleware.ts
```diff
@@ -61,6 +61,22 @@ const featureModuleMiddleware = [
   UserMiddleware.uploadUserPhotoFailureFlow({ notification: showSimpleMessage }),
   UserMiddleware.updateUserPhotoSuccessFlow({ notification: showSimpleMessage }),
   UserMiddleware.updateUserPhotoFailureFlow({ notification: showSimpleMessage }),
+  UserMiddleware.fetchUserCredentialsFlow,
+  UserMiddleware.fetchUserCredentialsSuccessFlow,
+  UserMiddleware.fetchUserCredentialsFailureFlow({ notification: showSimpleMessage }),
+  UserJobsMiddleware.createUserJobsFlow,
```
cameronolivier: I expect this will update once you've refactored the rest.

src/modules/UserJobs/UserJobs.utils.tsx
```diff
@@ -0,0 +1,23 @@
+import { UserCredentialTypes } from 'api/users/users.types'
+import { applySpec, path, pick, pipe } from 'ramda'
+
```
cameronolivier: It looks like none of these are required for this work

cameronolivier (changes requested):

src/modules/User/User.middleware.ts
sciphergfx: I'll have a look at your implementation. As discussed, i thought this was meant to pull the respective credentials in to their states

src/modules/UserJobs/UserJobs.middleware.tsx
sciphergfx: I extracted most of these changes to make the PR smaller

src/modules/UserJobs/UserJobs.types.tsx
sciphergfx: we have payload suffixed in other parts of the code, it was a reference but noted

src/modules/UserJobs/UserJobs.types.tsx
sciphergfx: It also serves as a type for an API response payload

src/utils/redux.utils.ts
sciphergfx: As per the API documentation, I added this to support credential responses that are in object shape after a successful create/update operation

src/modules/UserJobs/UserJobs.types.tsx
sciphergfx: I'll re-add this. some how landed in another PR

src/modules/Experience/Experience.types.ts
```diff
@@ -1,15 +1,8 @@
+import { UserJobsOrganisation } from 'modules/UserJobs/UserJobs.types'
+
 export type ExperienceType = {
   endDate: string
   id: string
-  job: Job
+  job: UserJobsOrganisation
```
cameronolivier: this still doesn't really make sense - is this for a UserJob or a UserJobOrganisation - and if the latter - why is the property called `job`?

src/utils/redux.utils.ts
cameronolivier: Ok.

cameronolivier: Please update the tests to cover this use case.

src/modules/UserJobs/UserJobs.types.tsx
cameronolivier: I hear you. they might need to be refactored.

src/modules/User/User.middleware.ts
cameronolivier: Maybe there was a misunderstanding - but I definitely said many times that each of the modules handle pulling the data they need out of the `fetchUserCredentialsSuccess` action payload.

src/modules/UserJobs/UserJobs.middleware.tsx
cameronolivier: I don't think we're on the same page. 

We're trying to achieve 2 things. 1 - we need to break the larger chunks of work into smaller chunks of work. but 2. those chunks still need to make sense with what we're trying to achieve. As this was said to be a 'add jobs data to state' task - it doesn't make sense to do all crud tasks here. If we are doing all CRUD tasks in this PR, then I'll review it as such - but I think it makes more sense to do each of the pieces of functionality as we need them. 

So in preparation for the displaying data (451), we need to store data.
In doing the 452 -handling edit/add/delete, we need to be able to update or add or delete via the api.
I suppose it's how we want to break the work up - but once again - if the story says you're doign something try stick to that thing.

src/modules/UserJobs/UserJobs.middleware.tsx
```diff
@@ -0,0 +1,290 @@
...
+  updateUserJobsCredentials,
+  updateUserJobsCredentialsFailure,
+  updateUserJobsCredentialsSuccess,
+  updateUserJobsFailure,
+  updateUserJobsSuccess,
+} from './UserJobs.reducer'
+import { selectUserJobsCredentialIdFromTmpFormValues, selectUserJobsTmpFormValues } from './UserJobs.selector'
+import { UserJobsCredentialsTmpFormValues } from './UserJobs.types'
+import {
+  extractUserJobsCredentialId,
```
cameronolivier: I don't think we need to user 'credentials' in our naming here.

I think `Jobs` for unnattached jobs and `UserJobs` for user experience is 100% - and the pattern we should use for the rest of the credentials work.

src/modules/UserJobs/UserJobs.middleware.tsx
```diff
@@ -0,0 +1,290 @@
...
+        ),
+      )
+
+      const tmpFormValues = extractUserJobsCredentialTmpFormValues(action)
+      dispatch(setTmpFormValues(tmpFormValues))
+    }
+    return result
+  }
+
+export const createUserJobsSuccessFlow: Middleware =
```
cameronolivier: this should be singular, right? we're creating a single job.

src/modules/UserJobs/UserJobs.middleware.tsx
```diff
@@ -0,0 +1,290 @@
...
+export const createUserJobsFlow: Middleware =
+  ({ dispatch }) =>
+  next =>
+  action => {
+    const result = next(action)
+
+    if (createUserJobs.match(action)) {
+      dispatch(
+        ApiActions.apiRequest(
+          mergeRight(ApiUserJobsConstants.JOBS_CREATE_CONFIG, {
```
cameronolivier: this is not ApiUserJobsConstants but simply ApiJobsConstants.

Please move all Jobs creation to the Jobs module as discussed.
All User Job Credentials CRUDS should be called UserJob(s) and be in the UserJobs module. As discussed.

cameronolivier: And again we're moving away from what was put in the job name. This is why I keep pushing to do specifically what we set our stories to - because when we do too much - each little change has knock-on effects. You know you can't properly create a new job experience credential without first creating a new job, and creating/tying to an organisation - and we'll be right back where we started with 452 - in being too large. It's frustrating for me that you don't get this and that there is constant push back on these things - which actually make us faster and make sure we're not diverging too far from what we're trying to achieve.

cameronolivier (changes requested):
Changes to date. Please complete these current changes before I'll review again.

src/redux/reducers.ts
```diff
@@ -1,12 +1,16 @@
...
 import { reducer as auth } from '../modules/Auth'
+import { reducer as job } from '../modules/Jobs'
 import { reducer as user } from '../modules/User'
 import { reducer as userChallenges } from '../modules/UserChallenges'
+import { reducer as userJobs } from '../modules/UserJobs'
 
 const rootReducer = combineReducers({
   auth,
   user,
+  job,
```
cameronolivier: issue: we need a consistent naming convention around plural/singular.
suggestion: [blocking] this needs to be 'jobs'

src/modules/UserJobs/UserJobs.reducer.test.ts
```diff
@@ -0,0 +1,74 @@
+import SUT, { clearUserJobs, INITIAL_STATE, setUserJobs } from './UserJobs.reducer'
+
+describe('src/modules/User/Jobs/Jobs.redux', () => {
```
cameronolivier: nitpick: [optional] Jobs.redux should be Jobs.reducer

src/modules/User/User.middleware.ts
cameronolivier: nitpick: please move this closer to `fetchUserCredentialsFlow`

cameronolivier (changes requested):
Can you bump the version number so long - we can then update if something else goes in first - but at least we won't miss it.

release-notes.txt
```diff
@@ -1,13 +1,13 @@
 Ver 0.0.1
 Latest:
-- adding User Challenges credentials to state
+- added jobs and user jobs to state
 
 Recent:
-- adding helper utils for handling MyCV data
 - docs upgrade, no new features added.
```
cameronolivier, line 8: nitpick: 
Please add the latest release notes under **Latest**
and move the current latest to top of the **Recent**

```text
Ver 0.0.1
Latest:
- added jobs and user jobs to state

Recent:
- adding User Challenges credentials to state
- adding helper utils for handling MyCV data
- docs upgrade, no new features added.
- Adjusted the widgets on MyCv to be able to take different types of content (previously they could only take text)
```

cameronolivier (changes requested):

cameronolivier (approved):

PR #112, opened 2021-08-12
YOMA-504: feat(UserChallenges): add real data to challenges widget

cameronolivier:
## Scope [YOMA-504](https://yomaworld.atlassian.net/browse/YOMA-504) of [YOMA-447](https://yomaworld.atlassian.net/browse/YOMA-447)
- adding real data into MyCV UserChallenges widget


src/modules/UserChallenges/Widget/UserChallengesWidget.container.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+import { useSelector } from 'react-redux'
+
+import selector from '../UserChallenges.selector'
+import UserChallengesWidget from './UserChallengesWidget'
+
+const UserChallengesWidgetContainer = () => {
```
cameronolivier: I know this doesn't follow our current container pattern - but I actually think it still does the same thing-  but also makes it a ton easier to type - and to pass in external props along with state-based props if need be (which is possible with the old way, but super tough to type)

src/modules/UserChallenges/Widget/UserChallengesWidget.selector.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export default createSelector(selectUserChallenges, challenges => {
+  const ids = slice(0, 2, challenges.ids)
+  const entities = pipe(
+    pick(ids),
+    map(
+      applySpec({
+        name: pathOr('', ['challenge', 'name']),
+        startDate: propOr('', 'startDate'),
+        avatarUrl: path(['challenge', 'url']),
```
sciphergfx: nitpick: https://github.com/didx-xyz/yoma-mobile/pull/116#discussion_r692868172

src/modules/Auth/Auth.selector.ts
```diff
@@ -2,7 +2,7 @@ import { createSelector } from '@reduxjs/toolkit'
 import { complement, isNil, path, pick, pipe, propOr } from 'ramda'
 import { RootState } from 'redux/redux.types'
 
-export const selectAuthState = (state: RootState) => state.auth
```
sciphergfx, line 5: suggestion(non-blocking): To keep PR file changes to the minimum, I think we can effect changes such at this in a separate ticket since it would be needed elsewhere on our code base (I think we have other instances where naming can be better). Also helps us standardise naming app-wide.

src/modules/UserChallenges/Widget/UserChallengesWidget.container.tsx
sciphergfx: I understand you, just so there's a flow, shall we restrict this selector pattern to widgets?

src/modules/Auth/Auth.selector.ts
cameronolivier: I disagree. This was a tiny refactor in 1-2 files of 16 files total. We should (within reason) always be updating/refactoring where it makes sense when we find them, while we work on things.

My rule of thumb is - small/medium unrelated refactors in up to medium size <20 file PR is 100% ok (Where it touches a handful of files)

A slightly bigger refactor would work on small PRs <7 files, but should generally be small - and ideally kept to the files you're working on.

Any proper refactor, obviously should be planned and in a separate story - this I agree with.

If we kept all refactors (especially as small as this) to separate PRs we'd never actually get to refactoring anything.

src/modules/UserChallenges/Widget/UserChallengesWidget.selector.ts
cameronolivier: Have updated.

src/modules/UserChallenges/Widget/UserChallengesWidget.container.tsx
cameronolivier: I think we can make use of it where it's required or makes sense. 

At the very least - you'll see - this makes sense in situations where we need to input props into the container for the inner component, along with selector & dispatch props. 

To be honest, I actually like this as a general pattern - it seems to be working quite well - and it means we can keep everything properly typed easier than with the `connect()` pattern. I'd argue it should be fine to use interchangeably as you wish - so long as we maintain the selectors in the selector file, I don't see it actually make much difference.

Happy to discuss further if need be.

sciphergfx (dismissed):

sciphergfx (approved):

PR #113, opened 2021-08-16
YOMA-506-feat(CredentialCertificate): add credential certificate data to state

sciphergfx:
## Work Done
- adds CredentialCertificate to state

Blocks 
- createCredentialItemSuccessFlow needs to pull updated credential type object (We're currently doing this using get<CredentialType>FromCredentialsFlow), need to agree on approach here.


src/redux/store.ts
```diff
@@ -11,6 +11,7 @@ const persistConfig = {
   key: 'root',
   version: 1,
   storage: AsyncStorage,
+  // blackList: ['credentialItems'],
```
sciphergfx, line 14: as per the API "credentialItems" is temporary created, not sure if we need this persisted?

cameronolivier (commented):
I'm going to hold off reviewing until I understand better what this is and why we need it.

PR #115, opened 2021-08-19
YOMA-533 - feat(UserJobs): Add user job credential to state

sciphergfx:
## Work Done
- Redux setup from userJobs creation
- Modified UserJobs State to hold temporary values for userJobs credential creation
- Added util to update state entities on when User<credentialType> is created
- Added tests where necessary


src/utils/redux.utils.ts
```diff
@@ -11,4 +11,11 @@ export const normalise = (data: Record<string, any>[], identifier = 'id') =>
     }),
   )(data)
 
+export const extractUpdatedNormalisedState = (
+  normalisedUpdate: Record<string, any>,
+  normalisedState: Record<string, any>,
+) => mergeDeepWith(concat, normalisedState, normalisedUpdate)
+
+export const extractNormalisedEntitiesFromState = pick(['ids', 'entities'])
```
cameronolivier: nitpick/suggestion: this is really just a selector - anything from state should use 'select' - so this is sort of a selector utility - so I'd suggest naming it `selectNormalised`.

src/modules/UserJobs/UserJobs.selector.tsx
```diff
@@ -0,0 +1,6 @@
+import { createSelector } from '@reduxjs/toolkit'
+import { path } from 'ramda'
+import { RootState } from 'redux/redux.types'
+
+export const selectUserJobs = (state: RootState) => state.userJobs
+export const selectUserJobsFormValues = createSelector(selectUserJobs, path(['formValues']))
```
cameronolivier: suggestion: `prop` would be better than `path` in this situation.

src/modules/UserJobs/UserJobs.reducer.ts
```diff
@@ -1,18 +1,31 @@
...
   ids: [],
   entities: {},
+  formValues: null,
 } as UserJobsState
 
 export const getUserJobsSuccess = createAction<UserJobCredential[]>(`${name} getUserJobsSuccess`)
 export const normaliseUserJobsSuccess = createAction<NormalisedUserJobs>(`${name} normaliseUserJobsSuccess`)
+export const updateNormalisedUserJobsState = createAction<NormalisedUserJobs>(`${name} updateNormalisedUserJobsState`)
+export const setUserJobsFormValues = createAction<UserJobsFormValues>(`${name} setUserJobsFormValues`)
+export const createUserJobs = createAction<any>(`${name} createUserJobs`)
```
cameronolivier: suggestion/nitpick: I presume this should be singular? `createUserJob`
nitpick: And could you type the payload.

src/modules/UserJobs/UserJobs.reducer.test.ts
```diff
@@ -2,7 +2,7 @@ import SUT, { clearUserJobs, INITIAL_STATE, setUserJobs } from './UserJobs.reduc
 
 describe('src/modules/User/Jobs/Jobs.reducer', () => {
   describe('setUserJobs', () => {
-    it('should correctly add the jobs credentials', () => {
+    it('should correctly add the usrJobs data', () => {
```
cameronolivier: nitpick: spelling mistake : should be userJobs or user jobs

src/modules/User/User.types.ts
```diff
@@ -63,5 +64,13 @@ export interface UserCredentialMeta {
   requestVerification: boolean
 }
 
+export interface UserCredentialItemPayload {
```
cameronolivier, line 68: question: is this the shape of the data returned when adding a credential?

src/modules/User/User.types.ts
```diff
@@ -63,5 +64,13 @@ export interface UserCredentialMeta {
...
+  type: UserCredentialTypes
+  credentialItemId: string
+  startTime: string
+  endTime: string
+  requestVerification: boolean
+}
+
 // TODO: Add Credentials as we type them as  (UserChallenge | UserExperience | ...)[]
 export type UserCredentials = UserChallenge[]
```
cameronolivier: issue: please add the UserJobs[] to this.

src/modules/User/User.utils.ts
```diff
@@ -25,3 +26,13 @@ export const createPhotoFormPayload = (formInstance: any) => (imageResponse: any
 }
 
 export const extractCredentialsByType = (type: UserCredentialTypes) => filter(pipe(keys, find(equals(toLower(type)))))
+
+export const extractCredentialPayload =
```
cameronolivier: issue: please add tests

src/modules/UserJobs/UserJobs.types.ts
```diff
@@ -1,3 +1,6 @@
+import { UserCredentialTypes } from 'api/users/users.types'
+import { ApiMetaResponse } from 'modules/Auth/Auth.types'
```
cameronolivier: suggestion - if this is a shared type now, we should refactor this and move it to a shared folder - ie if it's not Auth-specific. 

maybe in api.types in the src/types folder? or if you can think of a better place for that.

src/modules/UserJobs/UserJobs.types.ts
```diff
@@ -22,10 +25,23 @@ export interface UserJob {
...
+}
 
 export interface NormalisedUserJobs {
   ids: string[]
   entities: Record<string, UserJobCredential>
 }
 
-export interface UserJobsState extends NormalisedUserJobs {}
+export interface UserJobsState extends NormalisedUserJobs {
+  formValues?: UserJobsFormValues | null
```
cameronolivier: nitpick: this shouldn't be optional as it's always there - it's just either `UserJobsFormValues | null` so can drop the `?`

src/modules/UserJobs/UserJobs.utils.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { UserCredentialItemPayload } from 'modules/User/User.types'
+import { always, applySpec, merge, path } from 'ramda'
+import { StdFn } from 'types/general.types'
+
+import { UserJobsFormValues } from './UserJobs.types'
+
+export const extractJobId = path(['payload', 'id'])
+
+export const extractUserJobsPayload = (action: any): StdFn<any, UserCredentialItemPayload> =>
+  merge({
```
cameronolivier: issue: `merge` is deprecated - rather use `mergeRight`  or `mergeDeepRight` if you want recursive merge.

src/modules/UserJobs/UserJobs.utils.tsx
```diff
@@ -0,0 +1,23 @@
+import { UserCredentialTypes } from 'api/users/users.types'
+import { UserCredentialItemPayload } from 'modules/User/User.types'
+import { always, applySpec, merge, path } from 'ramda'
+import { StdFn } from 'types/general.types'
+
+import { UserJobsFormValues } from './UserJobs.types'
+
+export const extractJobId = path(['payload', 'id'])
```
cameronolivier: suggestions: it seems this is by its nature a generic function (it'll extract any id from a given payload. Might be useful to add it as a shared util in the `src/utils` folder - probably in api utils? or if you have a better idea as simply `extractId`

src/modules/UserJobs/UserJobs.utils.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+export const extractUserJobsPayload = (action: any): StdFn<any, UserCredentialItemPayload> =>
+  merge({
+    credentialItemId: extractJobId(action),
+  })
+
+export const extractUserCredentialFormValues = (
```
cameronolivier: nitpick: as this is generic, please move it to the User.utils file.

src/modules/UserJobs/UserJobs.utils.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+      // when ... we want to extract the id from the rest of the payload
+      const result = SUT.extractJobId(mockPayload)
+
+      // then ... the data should be extracted correctly
+      expect(result).toEqual('ID')
+    })
+  })
+  describe('extractUserCredentialFormValues', () => {
+    it('should return the jobs credentials form values from state', () => {
```
cameronolivier: question/nitpick: is this correct? aren't you getting the values from an action payload, not from state?

src/modules/UserJobs/UserJobs.utils.tsx
```diff
@@ -0,0 +1,23 @@
+import { UserCredentialTypes } from 'api/users/users.types'
+import { UserCredentialItemPayload } from 'modules/User/User.types'
+import { always, applySpec, merge, path } from 'ramda'
+import { StdFn } from 'types/general.types'
+
+import { UserJobsFormValues } from './UserJobs.types'
+
+export const extractJobId = path(['payload', 'id'])
+
+export const extractUserJobsPayload = (action: any): StdFn<any, UserCredentialItemPayload> =>
```
cameronolivier: suggestion: rename this to something that better describes what this is doing.

This function seems to be merging the create job payload with form data from state. 
Alternatively, it looks like it might be preparing the "create job credential" payload. 

So naming it so that it reflects what it's doing means it's easier to understand the code without having to interrogate the actual function. 

currently, our 'extract' utils are pulling data out of a payload. If we're doing more we should rename to reflect that.

src/modules/UserJobs/UserJobs.utils.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+        type: UserCredentialTypes.Job,
+        requestVerification: false,
+        startTime: 'START_TIME',
+        endTime: 'END_TIME',
+      })
+    })
+  })
+
+  describe('extractUserJobsPayload', () => {
+    it('should return the userJobs form values from state', () => {
```
cameronolivier: nitpick: I don't think this is what this is doing. because that's what the selector is doing. This is merging form values with payload data

src/utils/redux.utils.ts
```diff
@@ -11,4 +11,11 @@ export const normalise = (data: Record<string, any>[], identifier = 'id') =>
     }),
   )(data)
 
+export const extractUpdatedNormalisedState = (
```
cameronolivier: issue/nitpick:This isn't extracting anything - it shouldn't be prefixed with extract. it looks like a reducer utility, but it isn't extracting anything, so please rename. 

I'd suggest something like `updateNormalisedState`

src/modules/UserJobs/UserJobs.middleware.ts
```diff
@@ -46,3 +73,90 @@ export const setUserJobsFlow: Middleware =
...
+    return result
+  }
+
+export const updateNormalisedUserJobsStateFlow =
+  ({ notification }: { notification: typeof showSimpleMessage }): Middleware =>
+  ({ getState, dispatch }) =>
+  next =>
+  action => {
+    const result = next(action)
+    if (updateNormalisedUserJobsState.match(action)) {
```
cameronolivier: issue: what you're doing here should be handled in the reducer. 

suggestion: I'd suggest you do the following:

have an `updateUserJobs` action.
simply dispatch the payload of `normalisedJobs`

then in the reducer, use a version of your `extractUpdatedNormalisedState` (or as suggested `updateNormalisedState`) util.
```ts
builder.addCase(updateUserJobs, (state, action) => updateNormalisedState(state, action.payload))
```
and that's it. :)

src/modules/UserJobs/UserJobs.middleware.ts
```diff
@@ -46,3 +73,90 @@ export const setUserJobsFlow: Middleware =
...
+
+export const createUserJobsSuccessFlow: Middleware =
+  ({ dispatch }) =>
+  next =>
+  action => {
+    const result = next(action)
+    if (createUserJobsSuccess.match(action)) {
+      const response = extractDataFromPayload(action)
+      const normalisedJobs = normalise([response])
+      dispatch(updateNormalisedUserJobsState(normalisedJobs))
```
cameronolivier: suggestion: Here's a suggested middleware flow that will be a bit cleaner: 

`createUserJobSuccessFlow`: (note this should be singular `Job`
do: 
- extractDataFromPayload
- normalise the data.
- dispatch: updateUserJobs(normalisedJobs)
- Navigation.navigate(HomeNavigationRoutes.Experience)
- notification('success', 'Details saved!')

then catch `updateUserJobs` in the reducer and simply merge state there (I've pasted the relevant bits from my comment below in `updateNormalisedUserJobsStateFlow`):

In the reducer, use a version of your `extractUpdatedNormalisedState` (or as suggested `updateNormalisedState`) util.
```ts
builder.addCase(updateUserJobs, (state, action) => updateNormalisedState(state, action.payload))
```

src/modules/UserJobs/UserJobs.selector.tsx
cameronolivier: suggestion: I think you could shorten this to simply `selectFormValues` - you can see other selectors- you don't really need to name explicitly. If you do use this selector in a different module our pattern is to import the module as `UserJobsSelectors` - so usage would be `UserJobsSelectors.selectFormValues` which would be fine.

cameronolivier (changes requested):

src/modules/User/User.types.ts
sciphergfx: So this is the request payload when adding a credential. Response is UserChallenge[]

src/modules/UserJobs/UserJobs.utils.tsx
sciphergfx: Do you have name suggestions here?

src/modules/UserJobs/UserJobs.reducer.ts
```diff
@@ -10,11 +12,20 @@ export const INITIAL_STATE = {
 
 export const getUserJobsSuccess = createAction<UserJobCredential[]>(`${name} getUserJobsSuccess`)
 export const normaliseUserJobsSuccess = createAction<NormalisedUserJobs>(`${name} normaliseUserJobsSuccess`)
+export const updateNormalisedUserJobs = createAction<NormalisedUserJobs>(`${name} updateNormalisedUserJobs`)
```
cameronolivier: suggestion/nitpick: This shouldn't specify `normalised` - because _state_ is normalised - so there's no need to make the distinction here. You're simply updating the userJobs in state so `updateUserJobs` is perfectly fine. Just as you have `setUserJobs` (which is also normalised - but there's no need to specify it there either).

src/modules/UserJobs/UserJobs.reducer.ts
```diff
@@ -10,11 +12,20 @@ export const INITIAL_STATE = {
...
+export const setUserJobsFormValues = createAction<UserCredentialFormValues>(`${name} setUserJobsFormValues`)
+
+export const createUserJob = createAction<any>(`${name} createUserJob`)
 export const setUserJobs = createAction<NormalisedUserJobs>(`${name} setUserJobs`)
 export const clearUserJobs = createAction(`${name} clearUserJobs`)
 
+export const createUserJobsSuccess = createAction<UserJobsResponse>(`${name} createUserJobsSuccess`)
```
cameronolivier: nitpick: (non-blocking) considering I'm already nitpicking - success/failure of `createUserJob` should also be singular.

src/modules/UserJobs/UserJobs.middleware.ts
```diff
@@ -47,3 +69,76 @@ export const setUserJobsFlow: Middleware =
...
+  action => {
+    const result = next(action)
+    if (createJob.match(action)) {
+      const formValues = extractUserCredentialFormValues(UserCredentialTypes.Job)(action)
+      dispatch(setUserJobsFormValues(formValues))
+    }
+    return result
+  }
+
+export const createUserJobsFlow: Middleware =
```
cameronolivier: nitpick should also be singular.

src/modules/UserJobs/UserJobs.middleware.ts
```diff
@@ -1,10 +1,32 @@
+import { UserCredentialTypes } from 'api/users/users.types'
+import { extractErrorMessageFromPayload } from 'modules/Error/error.utils'
+import { HomeNavigationRoutes } from 'modules/HomeNavigation/HomeNavigation.types'
+import { createJob } from 'modules/Jobs/Jobs.reducer'
+import { selectId } from 'modules/User/User.selector'
```
cameronolivier: issue: this needs to be `import * as UserSelectors from 'modules/User/User.selector'`, because it gives context to what `selectId` is doing. Right now it's not obvious which id selectId is retrieving from state.

by importing as `* as UserSelectors` you can use it as: `UserSelectors.selectId` which now makes more sense. 

We should be importing all external module imports for these types of contexts in this manner.

src/modules/UserJobs/UserJobs.middleware.ts
```diff
@@ -47,3 +69,76 @@ export const setUserJobsFlow: Middleware =
...
+            onFailure: createUserJobsFailure,
+          }),
+          userJobsPayload,
+        ),
+      )
+    }
+    return result
+  }
+
+export const createUserJobsSuccessFlow =
```
cameronolivier: nitpick: (non-blocking) should be singular

src/modules/UserJobs/UserJobs.middleware.ts
```diff
@@ -47,3 +69,76 @@ export const setUserJobsFlow: Middleware =
...
+      dispatch(updateNormalisedUserJobs(normalisedJobs))
+      //TODO: add navigation as a dependency
+      Navigation.navigate(HomeNavigationRoutes.Experience)
+      // TODO: this should be handled by the notification module
+      notification('success', 'Details saved!')
+    }
+    return result
+  }
+
+export const createUserJobsFailureFlow =
```
cameronolivier: nitpick: (non-blocking): should be singular

src/modules/UserJobs/UserJobs.middleware.ts
```diff
@@ -1,10 +1,32 @@
...
+  createUserJob,
+  createUserJobsFailure,
+  createUserJobsSuccess,
+  getUserJobsSuccess,
+  normaliseUserJobsSuccess,
+  setUserJobs,
+  setUserJobsFormValues,
+  updateNormalisedUserJobs,
+} from './UserJobs.reducer'
+import { selectFormValues } from './UserJobs.selector'
```
cameronolivier, line 29: comment: it's fine to import within the module like this - just external modules should be `* as ...`

src/modules/Jobs/Jobs.middleware.tsx
```diff
@@ -1,13 +1,12 @@
-import { HomeNavigationRoutes } from 'modules/HomeNavigation/HomeNavigation.types'
+import { createUserJob } from 'modules/UserJobs/UserJobs.reducer'
```
cameronolivier: issue: this should be `import * as UserJobsActions` and used as `UserJobsActions.createUserJob`

cameronolivier (changes requested):

src/modules/UserJobs/UserJobs.reducer.ts
sciphergfx: this is named "updateNormalisedUserJobs" as "updateUserJobs" edit actions for API is coming in.

cameronolivier: Then we need a better naming convention.

Adding `normalised` doesn't add any value here or indicate whether it's updating state or the api. 

Using `select`/`set`/`update` for state naming makes sense.

I think we should adjust our naming convention for setting it to `adjust`.
`fetch`/`create`/`adjust` for api methods. 

What you think?

cameronolivier (approved):

src/modules/UserJobs/UserJobs.reducer.ts
sciphergfx: Outdated comment this seems, changes was already made here

PR #116, opened 2021-08-19
YOMA-537 feat(UserChallenges): update user challenges widget component

cameronolivier:
## Scope [YOMA-537](https://yomaworld.atlassian.net/browse/YOMA-537). Parent: [YOMA-447](https://yomaworld.atlassian.net/browse/YOMA-447)
- updating widget component, adding verified tick and avatar. 

## Work Done
- updated UserChallenges widget - moved to Widget folder
- adding Verified Tick component
- adding Avatar component

### NOTE:
I've added empty data in the current widget in the DigitalCV component - but this will be updated with the next PR that adds the data.

## Steps to Test
- run app - see MyCV page. Should have no data yet.
    
## Screenshots
Screenshot 2021-08-19 at 22 19 30
Screenshot 2021-08-19 at 23 07 40
Screenshot 2021-08-19 at 23 08 08
Screenshot 2021-08-19 at 23 23 58


src/modules/UserChallenges/UserChallenges.constants.ts
```diff
@@ -2,6 +2,7 @@ import { UserChallengeFormFields } from './UserChallenges.types'
 
 export const MOCKED_CHALLENGES = [
   {
+    id: 'id1',
     challenge: 'COVID Challenge',
     organisationLogoUrl: '',
     challengeHostProvider: '',
```
sciphergfx, line 8: question: is challengeHostProvider part of variables sent/consumed by the user as per API?

src/modules/DigitalCv/DigitalCv.tsx
```diff
@@ -59,14 +59,20 @@ const DigitalCv = ({ navigation, biography }: Props) => {
...
+        <UserChallengesWidget
+          navigation={navigation}
+          challenges={{
+            ids: ['id1'],
+            entities: {
+              id1: {
+                startDate: '2021-04-25T19:05:54.5496363',
+                isValidated: true,
+                name: 'ORG Verified with logo',
+                avatarUrl: 'https://picsum.photos/200',
```
sciphergfx: suggestion: I'm presuming this is an organisationLogoURL? In order to avoid confusion, I think retaining the correct prop names at the UI level where necessary may help us dispel potential confusions in the future.

I understand we have an avatar component that takes an avatarUrl, we can simply feed it "organisationLogoURL" as a prop, then the component points to it as "url" internally.

I think it should apply everywhere else if you're happy to effect this change.

src/modules/UserChallenges/UserChallenges.constants.ts
cameronolivier: I think this is old data - from when we first set this up. To be honest, I'm not sure - why you ask?

with Yoma-504, this will fall away for the widget - We should delete it when we do the Screen for challenges. I just added the id to make the widget work.

sciphergfx: So I found the same prop in src/modules/UserChallenges/UserChallenges.form.tsx as 
`<Input name={'challengeHostProvider'} label={t('Challenge host provider')} handlers={formikHandlers} />`
hence asking

src/modules/DigitalCv/DigitalCv.tsx
cameronolivier: Hey - so I partially agree. I agree -  I shouldn't have tied the prop to the avatar prop name. But also, don't think we need to follow the naming convention from the API either. Think we should stick to naming according to the requirements of the components themselves. That said - in this case that naming coincides with the api's :)

sciphergfx (approved):

src/modules/UserChallenges/UserChallenges.constants.ts
cameronolivier: yeah - it was old mocked data from when Ranjit was building out the UI before we had wired anything up - it should get removed as soon as possible :)

PR #117, opened 2021-08-19
YOMA-507 - feat(HomeHeader): add real data for profile image and zlto amount

cameronolivier:
## Scope [YOMA-507](https://yomaworld.atlassian.net/browse/YOMA-507)
- wire up zlto balance and profile image to use real data

## Work Done
- added selectors to User module for zlto balance and photo url
- adding default selector to HomeHeader to retrieve the required data
- adding container for HomeHeader
- moving HomeHeader into modules
- adjusting ProfilePhoto component so that real image is sized correctly - adding constant to manage size.

## Steps to Test
- open app - see header bar is pulling in correct info
    
## Screenshots
Screenshot 2021-08-20 at 00 31 23


sciphergfx (dismissed):
Tested, All good

sciphergfx (approved):

PR #118, opened 2021-08-22
YOMA-544 - feat(Skills): add filtered skills to state

sciphergfx:
## Work Done
- add filtered skills to state
- default number of skills to form component 20


src/modules/Skills/Skills.selector.test.tsx
```diff
@@ -0,0 +1,89 @@
...
+      })
+    })
+  })
+  describe('selectFiltered ', () => {
+    it('should return filtered property of the skills state', () => {
+      const stateMock = rootStateFixture({
+        skills: {
+          filtered: [
+            {
+              idA: 'Skill A',
```
cameronolivier: could you make these values different to the std values below - because in this test you could be pulling in the entities data and the test wouldn't warn you.

src/modules/Skills/Skills.selector.tsx
```diff
@@ -0,0 +1,7 @@
+import { createSelector } from '@reduxjs/toolkit'
+import { path } from 'ramda'
+import { RootState } from 'redux/redux.types'
+
+export const selectSkills = (state: RootState) => state.skills
+export const selectEntities = createSelector(selectSkills, path(['entities']))
```
cameronolivier: suggestion: prop is better than path here - it's simpler and to the point

src/modules/Skills/Skills.types.tsx
```diff
@@ -15,4 +15,6 @@ export type SkillsResponse = {
   meta: ApiMetaResponse
 }
 
-export interface SkillsState extends NormalisedSkills {}
+export interface SkillsState extends NormalisedSkills {
+  filtered?: []
```
cameronolivier: issue: this should be a typed array.

src/modules/Skills/Skills.utils.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+      expect(result).toEqual([
+        {
+          key: 'key1',
+          value: 'value1',
+        },
+      ])
+    })
+    it('should return 20 skills if value is empty', () => {
+      //given ...30 mocked skills list
+      const mockEntities = zipObj(keys(repeat(0, 30)), repeat({ key: 'key', value: 'skill' }, 30))
```
cameronolivier: suggestion: extract something like this into a little test helper util with a name that says what it does - which makes it easier to get what's happening in the test itself.

src/modules/Skills/Skills.reducer.tsx
```diff
@@ -1,23 +1,29 @@
...
 export const INITIAL_STATE = {
+  searchTerm: '',
   ids: [],
   entities: {},
 } as SkillsState
 
 export const setSkills = createAction<NormalisedSkills>(`${name} setSkills`)
 export const getSkillsSuccess = createAction<Skill[]>(`${name} getSkillsSuccess`)
 export const normaliseSkillsSuccess = createAction<NormalisedSkills>(`${name} normaliseSkillsSuccess`)
+export const setFilterSearchTerm = createAction<any>(`${name} setFilterSearchTerm`)
```
cameronolivier: nitpick - please type this - it should be a string I'd think

src/modules/Skills/Skills.utils.tsx
```diff
@@ -0,0 +1,15 @@
+import { curry, filter, isEmpty, mergeRight, pipe, slice, values } from 'ramda'
+
+import { SkillsState } from './Skills.types'
+
+//filter entity by search term else fallback to 20 records as default
+export const extractFilteredSkillsByValue = curry(({ searchTerm, entities }: SkillsState) => {
```
cameronolivier: suggestion/nitpick: this should be named `selectFiltered....` not `extractFiltered...` because you're selecting here. - or probably better:
`filterSkillsByValue` or `filterSkillsByTerm` <- I think this is best because you're filtering the skills by the search term.

src/modules/Skills/Skills.utils.tsx
```diff
@@ -0,0 +1,15 @@
...
+  const filtered = pipe(
+    values,
+    filter((obj: any) => obj.value.indexOf(searchTerm) === 0),
+  )(entities)
+
+  return isEmpty(searchTerm) ? slice(0, 20, filtered) : filtered
```
cameronolivier: Issue: we should exit before doing expensive work if we don't need to (ie do the searchTerm length/empty check first.

Then continue with the filtering only if it's not empty (ideally less than 3 length, correct?) 
(also, I don't think you need to curry this - unless you want to be able to call it as a curried function or as a single call.

Suggestion:
```ts
export const extractFilteredSkillsByValue = curry(({ searchTerm, entities, minTermLength = 3 }: SkillsState) => {

 if (searchTerm.length < minTermLength) {
    return slice(0, 20, entities)
 }

return pipe(
    values,
    filter((obj: any) => obj.value.indexOf(searchTerm) === 0),
  )(entities)

}

```

cameronolivier (changes requested):

src/modules/Skills/Skills.utils.tsx
sciphergfx: Ah.. I missed the minTermLength for this logic. With regards the function, I'm not sure we should be setting 'minTermLength' in redux state as suggested, if you noticed, {searchTerm, entities} were destructured. This approach will require we also convert entities to values for the first block which is more or less the same. (searchTerm.length < minTermLength) does negate the need for checking Empty values

cameronolivier: on the `minTermLength` I hear you - I think then store it as a constant - I suppose this isn't a function that would be used anywhere but in the set context - so that's fine.

Alternatively we can use `isEmpty` - it's easy enough to swap out (or simply set `minTermLength` to 0 and the check to `if (searchTerm.length <= minTermLength) {`.

Right. the code will be the same (So either extract it into a util that does that or simply replicate it) - but running that on 20 entities vs 2500++ (or however) will be vastly faster?.

Way to do it - is get id's in here too. Then simply slice first 20 ids, and map over that returning the value. something like this (ignore the naming): 
```ts
top20.map( id => entitites[id].value)
```

cameronolivier: suggestion: I wonder if the returned couldn't be refined a bit too. 

Ramda's `filter` function works directly on objects so you shouldn't need to run values on it first. If you want to return the values, I'd suggest swapping the 2 lines around:
```ts
return pipe(
    filter((obj: any) => obj.value.indexOf(searchTerm) === 0),
    values,
  )(entities)
```

nitpick: please type the obj.

src/modules/Skills/Skills.utils.tsx
```diff
@@ -0,0 +1,15 @@
+import { curry, filter, mergeRight, pipe, slice, values } from 'ramda'
+
+import { SkillsState } from './Skills.types'
+
+//filter entity by search term else fallback to 20 records as default
+export const selectFilterSkillsByTerm = curry(({ searchTerm = '', entities }: SkillsState, minTermLength = 3) => {
```
cameronolivier: suggestion: You don't need to curry this as it's a single argument (it just receives the skills state) so there's nothing to curry. (currying simply allows you to send `n` arguments to a function as single arguments across `n` functions).

_PS. Sorry I missed the destructuring in my previous comment - meant it as a separate argument._

cameronolivier (changes requested):
Left some further comments to the `selectFilterSkillsByTerm` util as comments (not part of this review) - so please make sure not to miss that :)

src/modules/Skills/Skills.utils.tsx
sciphergfx: just confirming no change to effect here?

src/modules/Skills/Skills.utils.tsx
sciphergfx: I'm a bit puzzled with the change required here. Do you have a code suggestion for this block perhaps? @cameronolivier

src/modules/Skills/Skills.utils.tsx
```diff
@@ -0,0 +1,18 @@
+import { curry, filter, map, mergeRight, pipe, slice, values } from 'ramda'
+
+import { FALLBACK_SKILLS_LENGTH, SEARCH_TERM_MIN_LENGTH } from './Skills.constants'
+import { SkillsState } from './Skills.types'
+
+export const selectFilterSkillsByTerm = curry(({ searchTerm = '', entities, ids }: SkillsState) => {
```
cameronolivier, line 6: nitpick: happy for you to refactor this in 452 or 451, but this doesn't need to be curry'ed

cameronolivier (dismissed):
Approved with comment.

src/modules/Skills/Skills.utils.test.tsx
```diff
@@ -0,0 +1,71 @@
+import { keys, repeat, zipObj } from 'ramda'
+
+import * as SUT from './Skills.utils'
+
+describe('modules/Skills/Skills.utils', () => {
+  describe('selectFilterSkillsByTerm', () => {
+    it('should extract skill name from entities', () => {
```
cameronolivier, line 7: suggestion: (non-blocking) I'd suggest adding a test to better show how this works - as in make the search term a partial match - but happy otherwise. Can do this in another PR if you wish.

cameronolivier (changes requested):
Hey. looks like you still need to bump the version

src/modules/Skills/Skills.utils.test.tsx
sciphergfx: need to tied all up on 452, I can effect this on the PR. I just need both PRs in if possible

cameronolivier (approved):

PR #119, opened 2021-08-26
YOMA-542 - refactor(UserCredentialItem): Create shared widget display item

sciphergfx:
## Work Done
- Support for shared widget item


src/modules/UserChallenges/Widget/UserChallengesWidget.tsx
```diff
@@ -3,9 +3,9 @@ import React from 'react'
 import { useTranslation } from 'react-i18next'
 
 import CvCard, { CvCardListBody } from '../../../components/CvCard'
+import * as UserChallengeItem from '../../../components/UserCredentialItem'
```
cameronolivier: issue: don't use `* as ...` please - it's a default export. 
also please call it `UserCredentialItem`.

cameronolivier (approved):
1 small thing. happy to do in the next work. 
Also - need to bump version and release notes... HOWEVER, considering the size, I'm ok with us not bumping things as long as we do it in the next one. 

I'm happy for you to make a call here.

src/modules/UserChallenges/Widget/UserChallengesWidget.tsx
sciphergfx: Miss this..Will effect the change

PR #121, opened 2021-08-27
YOMA-520 - feat(UserSkills): store user skills credentials in state

sciphergfx:
## Work Done
- redux setup for userSkills


src/modules/UserSkills/UserSkills.constants.ts
```diff
@@ -0,0 +1,25 @@
+export const FALLBACK_SKILLS_LENGTH = 20
+export const SEARCH_TERM_MIN_LENGTH = 3
+
+export const MOCK_SKILLS = [
```
cameronolivier, line 1: suggestion: we should delete the mock skills - I dont believe we need them anymore (do a usage check to confirm)

src/modules/Skills/Skills.constants.ts
```diff
@@ -1,25 +1,2 @@
 export const FALLBACK_SKILLS_LENGTH = 20
```
cameronolivier, line 1: issue: it seems we're repeating these two values - here and in the UserSkills constants

src/modules/UserSkills/UserSkillsForm/UserSkillsForm.constants.ts
```diff
@@ -1,10 +1,10 @@
-import { SkillsField } from './SkillsForm.types'
+import { UserSkillsField } from './UserSkillsForm.types'
 
-export const SKILLS_INITIAL_VALUES: SkillsField = {
+export const USER_SKILLS_INITIAL_VALUES: UserSkillsField = {
   skillNames: [],
 }
 
-export const MOCK_SKILLS_LIST = [
+export const MOCK_USER_SKILLS_LIST = [
```
cameronolivier, line 7: suggestion: we should delete this if we aren't using it (And we shouldn't be using it anymore).

src/modules/UserSkills/UserSkillsForm/UserSkillsForm.tsx
```diff
@@ -4,18 +4,18 @@ import React from 'react'
...
   const { t } = useTranslation()
 
   return (
-    <Formik initialValues={SKILLS_INITIAL_VALUES} onSubmit={() => {}}>
+    <Formik initialValues={USER_SKILLS_INITIAL_VALUES} onSubmit={() => {}}>
       {(formikHandlers: FormikProps<FormikValues>) => (
         <View style={styles.form}>
           <DropDownTags
-            items={MOCK_SKILLS_LIST}
+            items={MOCK_USER_SKILLS_LIST}
```
cameronolivier, line 18: we should be using real values going forward.

cameronolivier (changes requested):

src/modules/UserSkills/UserSkillsForm/UserSkillsForm.tsx
sciphergfx: considering the ticket is  storing user skills credentials in state don't think this is important

src/modules/UserSkills/UserSkillsForm/UserSkillsForm.constants.ts
sciphergfx: considering the ticket is storing user skills credentials in state don't think this is important.

src/modules/UserSkills/UserSkillsForm/UserSkillsForm.tsx
sciphergfx: there's a ticket that handles this

cameronolivier (approved):

